PR #7, opened 2019-02-14
Added const modifiers where it made sense.

JonathanHenson:


JonathanHenson:
const can't go into a struct like that.
1.) it will break C++ users since we aren't writing C++ constructors for them
2.) it will break us if declared initializers aren't used
3.) it will break non C99 customers that can't use declared initializers.

PR #38, opened 2020-03-24
Ecc key asn write

bretambrose:
Mac/Windows support can come later.


source/unix/opensslcrypto_ecc.c
```diff
@@ -383,3 +383,35 @@ struct aws_ecc_key_pair *aws_ecc_key_pair_new_from_asn1(
...
+    if (libcrypto_key_pair->ec_key == NULL) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    }
+
+    int encoded_length = i2d_ECPrivateKey(libcrypto_key_pair->ec_key, NULL);
+    if (encoded_length < 0) {
+        return aws_raise_error(AWS_ERROR_SYS_CALL_FAILURE);
+    }
+
+    if (aws_byte_buf_reserve_relative(buffer, (size_t)encoded_length)) {
```
JonathanHenson: Typically with these apis we have a mode to get length and then let the user make the allocation decisions

source/windows/bcrypt_ecc.c
```diff
@@ -470,3 +470,10 @@ struct aws_ecc_key_pair *aws_ecc_key_pair_new_from_asn1(
     }
     return NULL;
 }
+
+int aws_ecc_key_pair_append_asn1_encoding(struct aws_ecc_key_pair *key_pair, struct aws_byte_buf *buffer) {
+    (void)key_pair;
+    (void)buffer;
+
+    return AWS_OP_ERR;
+}
```
JonathanHenson: Newline.... just an FYI... you prolly just want BCRYPTExportKey when the time comes.

source/unix/opensslcrypto_ecc.c
```diff
@@ -383,3 +383,35 @@ struct aws_ecc_key_pair *aws_ecc_key_pair_new_from_asn1(
 
     return NULL;
 }
+
+int aws_ecc_key_pair_append_asn1_encoding(struct aws_ecc_key_pair *key_pair, struct aws_byte_buf *buffer) {
```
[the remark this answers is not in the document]
bretambrose: Maybe ..._export_der()?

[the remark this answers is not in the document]
bretambrose: Actually nm, we use asn1 elsewhere

source/unix/opensslcrypto_ecc.c
bretambrose: Took a while but makes sense; can guarantee only one allocation via byte_buf_init.

source/windows/bcrypt_ecc.c
bretambrose: Started down that route, it only seems to export in the internal windows format which isn't helpful.  Likely we'd need to just do the full DER encoding ourselves which would remove the platform nonsense but the time needed for impl/debugging did not seem worth it right now.

bretambrose:
Let's not do this after all.

PR #40, opened 2020-08-27
Disabled ecdsa key generation on iOS

justinboswell:
The APIs to export keys don't exist on iOS.

### DIFF
--- a/source/darwin/securityframework_ecc.c
+++ b/source/darwin/securityframework_ecc.c
@@ -3,12 +3,15 @@
  * SPDX-License-Identifier: Apache-2.0.
  */
 #include <Security/SecKey.h>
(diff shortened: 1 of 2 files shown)


source/darwin/securityframework_ecc.c
```diff
@@ -332,6 +335,7 @@ struct aws_ecc_key_pair *aws_ecc_key_pair_new_from_public_key(
     return NULL;
 }
 
+#if !defined(AWS_OS_IOS)
 struct aws_ecc_key_pair *aws_ecc_key_pair_new_generate_random(
```
graebm, line 339: should we ifdef out the declaration in the H file too?

justinboswell: YEs. Fixed.

PR #44, opened 2020-09-29
FindLibCrypto.cmake is easier to consume

graebm:
related to https://github.com/awslabs/aws-c-io/pull/323

Renamed FindLibCryptoCAL.cmake -> FindLibCrypto.cmake.
The unique name was to avoid collisions with the one installed by s2n, but now they install to different locations.

### DIFF
--- a/cmake/aws-c-cal-config.cmake
+++ b/cmake/aws-c-cal-config.cmake
@@ -3,7 +3,7 @@ include(CMakeFindDependencyMacro)
 find_dependency(aws-c-common)
 
 if (NOT BYO_CRYPTO AND NOT WIN32 AND NOT APPLE)
-    set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_LIST_DIR}/modules")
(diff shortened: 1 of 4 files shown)


cmake/aws-c-cal-config.cmake
```diff
@@ -1,7 +1,10 @@
 include(CMakeFindDependencyMacro)
 
+find_dependency(aws-c-common)
+
 if (NOT BYO_CRYPTO AND NOT WIN32 AND NOT APPLE)
-    find_dependency(LibCryptoCAL)
+    set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_LIST_DIR}/modules")
```
justinboswell, line 6: Same question as https://github.com/awslabs/aws-c-io/pull/323#discussion_r497159053

justinboswell (approved):
Fix?  and ship

cmake/aws-c-cal-config.cmake
graebm: yup, fixed while you were reviewing

PR #50, opened 2020-10-28
Dynamically determine Openssl contract via dlsym

bretambrose:
* Use dlsym to determine openssl api contract


source/unix/openssl_platform_init.c
```diff
@@ -0,0 +1,74 @@
...
+static struct openssl_evp_md_ctx_table evp_md_ctx_table;
+
+struct openssl_hmac_ctx_1_0_table *g_aws_openssl_hmac_ctx_1_0_table = NULL;
+struct openssl_hmac_ctx_1_1_table *g_aws_openssl_hmac_ctx_1_1_table = NULL;
+struct openssl_evp_md_ctx_table *g_aws_openssl_evp_md_ctx_table = NULL;
+
+void aws_cal_platform_init(struct aws_allocator *allocator) {
+    (void)allocator;
+
+    void *this_handle = dlopen(NULL, RTLD_NOW);
```
graebm, line 23: This is looking in the *current* DLL/EXE for the symbol.
What if we're building with `LibCrypto_SHARED_LIBRARY`? Wouldn't we want to look in `libcrypto.so` in that case?

source/unix/opensslcrypto_hmac.c
```diff
@@ -47,38 +81,19 @@ struct aws_hmac *aws_sha256_hmac_default_new(struct aws_allocator *allocator, co
         return NULL;
     }
 
-#if OPENSSL_VERSION_LESS_1_1
-    HMAC_CTX_init(ctx);
-#endif
+    if (g_aws_openssl_hmac_ctx_1_0_table != NULL) {
```
graebm, line 84: Could we just use [unsigned int OPENSSL_version_minor()](https://www.openssl.org/docs/manmaster/man3/OPENSSL_version_major.html) and avoid all this dlsym rigamarole?

source/unix/openssl_platform_init.c
bretambrose: Well not exactly, I believe the way it's set up now, it will look in the process space, which includes everything that has been loaded up to that point.  It's not a complete test (libcrypto is still static) but the lookups succeed when building as shared, so aws-c-cal-tests  is separate from lib-aws-c-cal.so

From dlopen:

"If the object specified by filename has dependencies on other shared
       objects, then these are also automatically loaded by the dynamic
       linker using the same rules.  (This process may occur recursively, if
       those objects in turn have dependencies, and so on.)"

I'm assuming the same is true when using NULL to go to the main program.

From dlsym's man:

" If the symbol is not found, in the specified library or any of the libraries that were automatically loaded by dlopen() when that library was loaded, dlsym() returns NULL. (The search performed by dlsym() is breadth first through the dependency tree of these libraries.)"

source/unix/opensslcrypto_hmac.c
bretambrose: That wouldn't help with the conditional compilation problem since the public contract/functions change.

graebm: If we're supporting both openssl 1.0 and 1.1, can we just call the deprecated 1.0 functions, and not call any of the 1.1-exclusive functions? when we stop supporting openssl 1.0 we can remove the deprecated calls and just call the new 1.1 functions.

I'm not familiar with what any of these functions do. I'm just hoping that the openssl authors are respecting the major version and not breaking 1.0-style code

justinboswell: They're not deprecated, they're GONE, unfortunately.

justinboswell: it's a fully breaking API change, requires conditional handling.

source/unix/openssl_platform_init.c
justinboswell: Bret is correct, this looks in the process space, including all loaded dynamic libs.

justinboswell (approved):

PR #53, opened 2020-11-17
vtable all the HMAC things

justinboswell:

### DIFF
--- a/source/unix/openssl_platform_init.c
+++ b/source/unix/openssl_platform_init.c
@@ -35,17 +35,17 @@ void aws_cal_platform_init(struct aws_allocator *allocator) {
     *(void **)(&reset_fn) = dlsym(this_handle, "HMAC_CTX_reset");
     *(void **)(&free_fn) = dlsym(this_handle, "HMAC_CTX_free");
 
-    if (init_fn != NULL && clean_up_fn != NULL) {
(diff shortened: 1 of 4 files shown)


source/unix/openssl_platform_init.c
```diff
@@ -9,43 +9,119 @@
...
+    AWS_PRECONDITION(
+        g_aws_openssl_hmac_ctx_table->init_fn != s_hmac_ctx_init_noop &&
+        g_aws_openssl_hmac_ctx_table->clean_up_fn != s_hmac_ctx_clean_up_noop &&
+        "libcrypto 1.0 reset called on libcrypto 1.1 vtable");
+    g_aws_openssl_hmac_ctx_table->clean_up_fn(ctx);
+    g_aws_openssl_hmac_ctx_table->init_fn(ctx);
+}
+
+void *s_find_libcrypto_module(void) {
+#if defined(AWS_CAL_EXPORTS)
```
bretambrose, line 61: Isn't this guaranteed always true or false since this is a .c file and so it's only the internal view?

bretambrose (approved):

source/unix/openssl_platform_init.c
justinboswell: It is only defined if we are building via `BUILD_SHARED=ON`

PR #56, opened 2020-11-19
ported crypto locking/id callbacks from aws-c-io

justinboswell:

### DIFF
--- a/source/unix/openssl_platform_init.c
+++ b/source/unix/openssl_platform_init.c
@@ -288,7 +288,8 @@ void aws_cal_platform_init(struct aws_allocator *allocator) {
 }
 
 void aws_cal_platform_clean_up(void) {
-    if (CRYPTO_get_locking_callback() == s_locking_fn) {
+    if (CRYPTO_get_locking_callback() != NULL &&
+        CRYPTO_get_locking_callback() == s_locking_fn) {
         CRYPTO_set_locking_callback(NULL);
         size_t lock_count = (size_t)CRYPTO_num_locks();
         for (size_t i = 0; i < lock_count; ++i) {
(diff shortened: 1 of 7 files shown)


source/unix/openssl_platform_init.c
```diff
@@ -92,24 +94,55 @@ static int s_hmac_ctx_reset(HMAC_CTX *ctx) {
...
+}
+
+static unsigned long s_id_fn(void) {
+    return (unsigned long)aws_thread_current_thread_id();
+}
+
+enum aws_libcrypto_version {
+    AWS_LIBCRYPTO_NONE,
+    AWS_LIBCRYPTO_102,
+    AWS_LIBCRYPTO_111,
```
graebm, line 118: this is confusing me.
how is "102" == "1.0.0"?
in any case it would probably help my understanding if we separated major/minor/patch like `AWS_LIBCRYPTO_1_1_1`

source/unix/openssl_platform_init.c
```diff
@@ -92,24 +94,55 @@ static int s_hmac_ctx_reset(HMAC_CTX *ctx) {
...
+    AWS_LIBCRYPTO_LC,
+} s_libcrypto_version = AWS_LIBCRYPTO_NONE;
+
 void *s_find_libcrypto_module(void) {
 #if defined(AWS_CAL_EXPORTS)
     const char *libcrypto_110 = "libcrypto.so.1.1";
     const char *libcrypto_102 = "libcrypto.so.1.0.0";
     void *module = dlopen(libcrypto_102, RTLD_NOW);
     if (module) {
+        s_libcrypto_version = AWS_LIBCRYPTO_102;
```
graebm, line 128: please comment  why we prefer the lower-versioned .so

graebm (approved):
fix & ship

source/unix/openssl_platform_init.c
justinboswell: 1.0.0 in the so-name is the ABI revision, not the software version. I'll fix the enum identifiers though.

source/unix/openssl_platform_init.c
justinboswell: can do

Issue #61, opened 2020-11-24
aws-c-cal 0.4.x is not compatible with aws-lc


petreeftime:
0.3.x version of aws-c-cal work with aws-lc (https://github.com/awslabs/aws-lc), but versions 0.4.x fail either at runtime with:
```
Fatal error condition occurred in ../source/unix/openssl_platform_init.c:100: update_fn != NULL && "libcrypto HMAC_Update could not be resolved"
```
or during compilation
```
[3/20] Building C object CMakeFiles/aws-c-cal.dir/source/unix/openssl_platform_init.c.o
FAILED: CMakeFiles/aws-c-cal.dir/source/unix/openssl_platform_init.c.o
/usr/bin/cc -DDEBUG_BUILD -DHAVE_SYSCONF -I../include -Wall -Werror -Wstrict-prototypes -pedantic -Wno-long-long -fPIC -pthread -std=gnu99 -MD -MT CMakeFiles/aws-c-cal.dir/source/unix/openssl_platform_init.c.o -MF CMakeFiles/aws-c-cal.dir/source/unix/openssl_platform_init.c.o.d -o CMakeFiles/aws-c-cal.dir/source/unix/openssl_platform_init.c.o   -c ../source/unix/openssl_platform_init.c
../source/unix/openssl_platform_init.c:37:12: error: conflicting types for 'HMAC_CTX_reset'
 extern int HMAC_CTX_reset(HMAC_CTX *) __attribute__((weak)) __attribute__((used));
            ^~~~~~~~~~~~~~
In file included from ../include/aws/cal/private/opensslcrypto_common.h:6:0,
                 from ../source/unix/openssl_platform_init.c:12:
/usr/include/openssl/hmac.h:142:21: note: previous declaration of 'HMAC_CTX_reset' was here
 OPENSSL_EXPORT void HMAC_CTX_reset(HMAC_CTX *ctx);
                     ^~~~~~~~~~~~~~
../source/unix/openssl_platform_init.c:46:12: error: conflicting types for 'HMAC_Init_ex'
 extern int HMAC_Init_ex(HMAC_CTX *, const void *, int, const EVP_MD *, ENGINE *) __attribute__((weak))
            ^~~~~~~~~~~~
In file included from ../include/aws/cal/private/opensslcrypto_common.h:6:0,
                 from ../source/unix/openssl_platform_init.c:12:
[...]

PR #65, opened 2021-02-10
Make Unix platforms build with AWS-LC

andrew-kaufman:
*Issue #, if available:* #61 

*Description of changes:* This makes Unix platforms build only with AWS-LC and removes support for older versions of OpenSSL's libcrypto. 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,8 +7,8 @@ on:
       - '!main'
 
 env:
-  BUILDER_VERSION: aws-lc
-  BUILDER_SOURCE: channels
+  BUILDER_VERSION: v0.8.2
+  BUILDER_SOURCE: releases
   BUILDER_HOST: https://d19elf31gohf1l.cloudfront.net
   PACKAGE_NAME: aws-c-cal
   LINUX_BASE_IMAGE: ubuntu-16-x64
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
         echo "${{ secrets.GITHUB_TOKEN }}" | docker login docker.pkg.github.com -u awslabs --password-stdin
         export DOCKER_IMAGE=docker.pkg.github.com/awslabs/aws-crt-builder/aws-crt-${{ matrix.image }}:${{ env.BUILDER_VERSION }}
         docker pull $DOCKER_IMAGE
-        docker run --env CFLAGS=-D__STDC_FORMAT_MACROS --env GITHUB_REF $DOCKER_IMAGE --version=${{ env.BUILDER_VERSION }}  build -p ${{ env.PACKAGE_NAME }} --spec=downstream
+        docker run --env GITHUB_REF $DOCKER_IMAGE --version=${{ env.BUILDER_VERSION }}  build -p ${{ env.PACKAGE_NAME }} --spec=downstream
 
   al2:
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


justinboswell (approved):

CMakeLists.txt
```diff
@@ -78,8 +80,8 @@ else ()
         file(GLOB AWS_CAL_OS_SRC
             "source/unix/*.c"
         )
-        find_package(LibCrypto REQUIRED)
-        set(PLATFORM_LIBS LibCrypto::Crypto dl)
+        find_library(AWS_LC crypto REQUIRED)
+        set(PLATFORM_LIBS dl ${AWS_LC})
```
graebm, line 84: trivial: What's more normal for Cmake? To define a variable, or just have a normal fixed string?

.github/workflows/ci.yml
```diff
@@ -79,7 +81,7 @@ jobs:
         echo "${{ secrets.GITHUB_TOKEN  }}" | docker login docker.pkg.github.com -u awslabs --password-stdin
         export DOCKER_IMAGE=docker.pkg.github.com/awslabs/aws-crt-builder/aws-crt-${{ env.LINUX_BASE_IMAGE  }}:${{ env.BUILDER_VERSION  }}
         docker pull $DOCKER_IMAGE
-        docker run --env GITHUB_REF $DOCKER_IMAGE --version=${{ env.BUILDER_VERSION  }} build -p ${{ env.PACKAGE_NAME  }} --compiler=gcc-8 --spec=downstream cmake_args='-DBUILD_SHARED=ON'
+        docker run --env GITHUB_REF $DOCKER_IMAGE --version=${{ env.BUILDER_VERSION  }} build -p ${{ env.PACKAGE_NAME  }} --compiler=gcc-8 --spec=downstream 'cmake_args=-DBUILD_SHARED_LIBS=ON'
```
graebm, line 84: I don't think the builder is handling cmake_args correctly yet.
When I "fixed" it the other day, I'd only tested in aws-c-common (which has no upstream dependencies).
When I tried it in aws-c-io, it didn't work.
I think you're hitting the same issue, where aws-c-common isn't getting built with -DBUILD_SHARED_LIBS=ON

CMakeLists.txt
justinboswell: Well, these are different. One is finding a package and a named target, one is finding a library and storing it in a variable, and then the expanded path is added to the link line.

CMakeLists.txt
```diff
@@ -24,6 +24,8 @@ elseif(NOT DEFINED CMAKE_INSTALL_LIBDIR)
     set(CMAKE_INSTALL_LIBDIR "lib")
 endif()
 
+set(FIND_LIBRARY_USE_LIB64_PATHS ON)
```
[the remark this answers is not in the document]
justinboswell: Oh good call, this is the result of some flailing. I'll pull it out and make sure we're good.

.github/workflows/ci.yml
justinboswell: yes, correct. We should probably just add a `--shared` to builder that `CMakeBuild` picks up

CMakeLists.txt
graebm: Sorry, I was not clear. I meant: is it normal for a find_library() call to have the side-effect of setting a variable `${AWS_LC}` that you then pass to PLATFORM_LIBS? Or is it more normal to just have the user type out both strings
```
find_library(AWS_LC crypto REQUIRED)
set(PLATFORM_LIBS dl LibCrypto::Crypto) # or whatever ${AWS_LC} expands to
```

graebm (approved):
fix and ship

PR #68, opened 2021-03-03
Do not call platform init with BYO_CRYPTO, Note, we still need to go back and add the hooks for ECC, but doing this alone is still an improvement.

JonathanHenson:
Do not call platform init with BYO_CRYPTO, Note, we still need to go back and add the hooks for ECC, but doing this alone is still an improvement.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
       - '!main'
 
 env:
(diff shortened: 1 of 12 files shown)


rccarper (approved):
Is it worth doing a null implementation that you can always fall back on instead of the macros?

justinboswell (approved):

PR #79, opened 2021-04-07
CRC32, CRC32C Integration and Testing

ilevyor:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cmake/aws-c-cal-config.cmake
+++ b/cmake/aws-c-cal-config.cmake
@@ -1,6 +1,7 @@
 include(CMakeFindDependencyMacro)
 
 find_dependency(aws-c-common)
+find_dependency(aws-checksums)
 
 if (NOT BYO_CRYPTO AND NOT WIN32 AND NOT APPLE)
     list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/modules")
--- a/include/aws/cal/hash.h
+++ b/include/aws/cal/hash.h
@@ -42,6 +42,10 @@ AWS_CAL_API struct aws_hash *aws_sha256_new(struct aws_allocator *allocator);
  */
 AWS_CAL_API struct aws_hash *aws_md5_new(struct aws_allocator *allocator);
 /**
+ * Allocates and initializes a crc32 hash instance
+ */
+AWS_CAL_API struct aws_hash *aws_crc32_new(struct aws_allocator *allocator);
+/**
  * Cleans up and deallocates hash.
  */
 AWS_CAL_API void aws_hash_destroy(struct aws_hash *hash);
--- a/include/aws/cal/hash.h
+++ b/include/aws/cal/hash.h
@@ -11,6 +11,8 @@
 
 #define AWS_SHA256_LEN 32
 #define AWS_MD5_LEN 16
+#define AWS_CRC32_LEN 4
+#define AWS_CRC32C_LEN 4
 
 struct aws_hash;
 
--- a/include/aws/cal/hash.h
+++ b/include/aws/cal/hash.h
@@ -48,6 +48,11 @@ AWS_CAL_API struct aws_hash *aws_md5_new(struct aws_allocator *allocator);
  */
 AWS_CAL_API struct aws_hash *aws_crc32_new(struct aws_allocator *allocator);
 /**
+ * Allocates and initializes a crc32c hash instance
+ */
+AWS_CAL_API struct aws_hash *aws_crc32c_new(struct aws_allocator *allocator);
+
+/**
  * Cleans up and deallocates hash.
  */
 AWS_CAL_API void aws_hash_destroy(struct aws_hash *hash);
--- a/source/crc32.c
+++ b/source/crc32.c
@@ -8,7 +8,7 @@ static int s_crc32_update(struct aws_hash *hash, const struct aws_byte_cursor *t
 static int s_crc32c_update(struct aws_hash *hash, const struct aws_byte_cursor *to_hash);
 static int s_finalize(struct aws_hash *hash, struct aws_byte_buf *output);
 static void s_destroy(struct aws_hash *hash) {
-    (void)hash;
+    aws_mem_release(hash->allocator, hash);
 }
 
 static struct aws_hash_vtable crc32_vtable = {
@@ -63,6 +63,7 @@ static int s_crc32_common_update(
     uintptr_t crc_value = (uintptr_t)hash->impl;
     uint32_t crc = (uint32_t)crc_value;
 
+    /* hash->impl has type (void *) to match the rest of the API, but we are storing a uintptr, and useing it as such */
     uintptr_t new_crc = aws_checksums(to_hash->ptr, (int)to_hash->len, crc);
     hash->impl = (void *)new_crc;
     return AWS_OP_SUCCESS;
--- a/source/crc32.c
+++ b/source/crc32.c
@@ -4,29 +4,27 @@
 #include <aws/common/byte_buf.h>
 #include <stdint.h>
 
-static int crc32_update(struct aws_hash *hash, const struct aws_byte_cursor *to_hash);
-static int crc32c_update(struct aws_hash *hash, const struct aws_byte_cursor *to_hash);
+static int s_crc32_update(struct aws_hash *hash, const struct aws_byte_cursor *to_hash);
+static int s_crc32c_update(struct aws_hash *hash, const struct aws_byte_cursor *to_hash);
 static int s_finalize(struct aws_hash *hash, struct aws_byte_buf *output);
 static void s_destroy(struct aws_hash *hash) {
     (void)hash;
 }
-/* can we reuse some of these as defaults defined in hash.c?*/
 
-/* is it possible to define this in a function somehow? I don't like that we define almost the same vtable twice */
 static struct aws_hash_vtable crc32_vtable = {
     .destroy = s_destroy,
-    .update = crc32_update,
+    .update = s_crc32_update,
     .finalize = s_finalize,
     .alg_name = "CRC32",
-    .provider = "AWS", /* is this right? */
+    .provider = "AWS",
 };
 
 static struct aws_hash_vtable crc32c_vtable = {
     .destroy = s_destroy,
-    .update = crc32c_update,
+    .update = s_crc32c_update,
     .finalize = s_finalize,
     .alg_name = "CRC32C",
-    .provider = "AWS", /* is this right? */
+    .provider = "AWS",
 };
 
 static struct aws_hash *s_crc32_common_init(struct aws_allocator *allocator, struct aws_hash_vtable *vtable) {
@@ -53,9 +51,11 @@ struct aws_hash *aws_crc32c_default_new(struct aws_allocator *allocator) {
     return s_crc32_common_init(allocator, &crc32c_vtable);
 }
 
-/* would it be useful to collect most of the logic into a helper function with a flag, and have two versions with the
-   correct paramters pass the corresponding flag? */
-static int crc32_update(struct aws_hash *hash, const struct aws_byte_cursor *to_hash) {
+static int s_crc32_common_update(
+    struct aws_hash *hash,
+    const struct aws_byte_cursor *to_hash,
+    uint32_t (*aws_checksums)(const uint8_t *, int, uint32_t)) {
+
     if (!hash->good) {
         return aws_raise_error(AWS_ERROR_INVALID_STATE);
     }
@@ -63,22 +63,17 @@ static int crc32_update(struct aws_hash *hash, const struct aws_byte_cursor *to_
     uintptr_t crc_value = (uintptr_t)hash->impl;
     uint32_t crc = (uint32_t)crc_value;
 
-    uintptr_t new_crc = aws_checksums_crc32(to_hash->ptr, (int)to_hash->len, crc);
+    uintptr_t new_crc = aws_checksums(to_hash->ptr, (int)to_hash->len, crc);
     hash->impl = (void *)new_crc;
     return AWS_OP_SUCCESS;
 }
 
-static int crc32c_update(struct aws_hash *hash, const struct aws_byte_cursor *to_hash) {
-    if (!hash->good) {
-        return aws_raise_error(AWS_ERROR_INVALID_STATE);
-    }
+static int s_crc32_update(struct aws_hash *hash, const struct aws_byte_cursor *to_hash) {
+    return s_crc32_common_update(hash, to_hash, &aws_checksums_crc32);
+}
 
-    uintptr_t crc_value = (uintptr_t)hash->impl;
-    uint32_t crc = (uint32_t)crc_value;
-
-    uintptr_t new_crc = aws_checksums_crc32c(to_hash->ptr, (int)to_hash->len, crc);
-    hash->impl = (void *)new_crc;
-    return AWS_OP_SUCCESS;
+static int s_crc32c_update(struct aws_hash *hash, const struct aws_byte_cursor *to_hash) {
+    return s_crc32_common_update(hash, to_hash, &aws_checksums_crc32c);
 }
 
 static int s_finalize(struct aws_hash *hash, struct aws_byte_buf *output) {
--- a/source/crc32.c
+++ b/source/crc32.c
@@ -19,7 +19,6 @@ static struct aws_hash_vtable s_vtable = {
     .provider = "AWS", /* is this right? */
 };
 
-
 struct aws_hash *aws_crc32_default_new(struct aws_allocator *allocator) {
     struct aws_hash *hash = aws_mem_acquire(allocator, sizeof(struct aws_hash));
 
@@ -30,7 +29,7 @@ struct aws_hash *aws_crc32_default_new(struct aws_allocator *allocator) {
     hash->allocator = allocator;
     hash->vtable = &s_vtable; /*could make to vtables, and pass an argument to default new*/
     hash->impl = 0;
-    hash->digest_size = AWS_CRC32C_LEN; 
+    hash->digest_size = AWS_CRC32C_LEN;
     hash->good = true;
 
     return hash;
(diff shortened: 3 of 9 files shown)


source/crc32.c
```diff
@@ -0,0 +1,73 @@
...
+/* hash is object, vtable is methods, hash properties (not vtable) are data?*/
+
+static void s_destroy(struct aws_hash *hash);
+static int s_update(struct aws_hash *hash, const struct aws_byte_cursor *to_hash);
+static int s_finalize(struct aws_hash *hash, struct aws_byte_buf *output);
+/* why do we need to define and implement the functions separatly? */
+/* can we reuse some of these as defaults defined in hash.c?*/
+
+/* does this have to be defined globaly? I'd like to use a different update function depending on a parameter */
+static struct aws_hash_vtable s_vtable = {
```
justinboswell, line 14: go ahead and make this s_crc32_vtable, we know that you're going to add s_crc32c_vtable next, and they share a lot.

source/crc32.c
```diff
@@ -0,0 +1,73 @@
...
+/* reordered from destroy update finalize to update finalize destroy, this order makes more sense to me */
+static int s_update(struct aws_hash *hash, const struct aws_byte_cursor *to_hash) {
+    if (!hash->good) {
+        return aws_raise_error(AWS_ERROR_INVALID_STATE);
+    }
+
+    uint32_t crc = (uint32_t)hash->impl;
+
+    /* should checksums be refactored to use a byte cursor instead? */
+    *((uintptr_t*)&(hash->impl)) = aws_checksums_crc32(to_hash->ptr, to_hash->len, crc);
```
justinboswell, line 48: Another, possibly easier/more-readable way to do this:
```c
uintptr_t new_crc = aws_checksums_crc32(...);
hash->impl = (void*)new_crc;
```

source/crc32.c
```diff
@@ -0,0 +1,73 @@
...
+        return aws_raise_error(AWS_ERROR_INVALID_STATE);
+    }
+
+    size_t buffer_len = output->capacity - output->len;
+
+    if (buffer_len < AWS_CRC32C_LEN) {
+        return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
+    }
+
+    /* CC_SHA256_Final(output->buffer + output->len, &ctx->cc_hash); crc doesn't need this line because the intermediat
```
justinboswell, line 63: Can delete this line.

source/crc32.c
```diff
@@ -0,0 +1,73 @@
...
+       is the same as the final correct? How do I follow function to implementaiton?*/
```
justinboswell, line 64: You'll need to do
```c
hash->good = false;
uint32_t crc = (uint32_t)hash->impl;
return aws_byte_buf_write_uint32(output, crc);
```

and get rid of the `output->len` update, write does it for you

source/crc32.c
```diff
@@ -0,0 +1,73 @@
...
+    hash->good = false;
+    output->len += buffer_len;
+    return AWS_OP_SUCCESS;
+}
+
+static void s_destroy(struct aws_hash *hash) {
+    /* uint32_t ctx = hash->impl; */
+    aws_mem_release(hash->allocator, hash->impl);
```
justinboswell, line 72: Don't need this, this function is a no-op for the crcs

source/hash.c
```diff
@@ -28,6 +29,12 @@ struct aws_hash *aws_md5_new(struct aws_allocator *allocator) {
     return s_md5_new_fn(allocator);
 }
 
+/* should this be in this file since all I'm doing is calling the other function, or should the other functions
+   implementation be here? */
+struct aws_hash *aws_crc32_new(struct aws_allocator *allocator) {
```
justinboswell, line 34: This is fine, this lets us separate things in case we need to later.

justinboswell (changes requested):

source/crc32.c
```diff
@@ -0,0 +1,72 @@
+#include <aws/cal/crc32.h>
+#include <aws/cal/hash.h>
+#include <aws/checksums/crc.h>
+
+/* hash is object, vtable is methods, hash properties (not vtable) are data?*/
+
+static void s_destroy(struct aws_hash *hash);
+static int s_update(struct aws_hash *hash, const struct aws_byte_cursor *to_hash);
+static int s_finalize(struct aws_hash *hash, struct aws_byte_buf *output);
+/* why do we need to define and implement the functions separatly? */
```
justinboswell, line 10: This is a forward declaration. Without it, the vtable code below doesn't know what these functions are. Another way to write this is to define those functions here, then declare the vtable below them in the file. Either way is fine, we use both interchangeably

source/crc32.c
```diff
@@ -0,0 +1,72 @@
...
+/* can we reuse some of these as defaults defined in hash.c?*/
```
justinboswell, line 11: no because they're declared static, which keeps them only visible to the translation unit they're in (e.g. their C file)

source/crc32.c
```diff
@@ -0,0 +1,72 @@
...
+    hash->allocator = allocator;
+    hash->vtable = &s_vtable; /*could make to vtables, and pass an argument to default new*/
+    hash->impl = 0;
+    hash->digest_size = AWS_CRC32C_LEN;
+    hash->good = true;
+
+    return hash;
+}
+
+/* reordered from destroy update finalize to update finalize destroy, this order makes more sense to me */
```
justinboswell, line 38: This comment doesn't match the code, can probably go?

source/crc32.c
```diff
@@ -0,0 +1,72 @@
...
+static int s_update(struct aws_hash *hash, const struct aws_byte_cursor *to_hash) {
+    if (!hash->good) {
+        return aws_raise_error(AWS_ERROR_INVALID_STATE);
+    }
+
+    uint32_t crc = (uint32_t)hash->impl;
+
+    /* should checksums be refactored to use a byte cursor instead? */
```
justinboswell, line 46: Nope. This is fine.

source/crc32.c
```diff
@@ -0,0 +1,72 @@
...
+
+struct aws_hash *aws_crc32_default_new(struct aws_allocator *allocator) {
+    struct aws_hash *hash = aws_mem_acquire(allocator, sizeof(struct aws_hash));
+
+    if (!hash) {
+        return NULL;
+    }
+
+    hash->allocator = allocator;
+    hash->vtable = &s_vtable; /*could make to vtables, and pass an argument to default new*/
```
justinboswell, line 30: You could indeed make an init function and put this in its contents:
```C
static struct aws_hash *s_crc32_common_init(struct aws_allocator *allocator, struct aws_hash_vtable *vtable);
```

source/crc32.c
ilevyor: This is in relation to the order the methods show up in Sha256CC. It doesn't really matter, but I think update finalize destroy makes more sense since I imagine that's the order the functions would be invoked during the lifetime of the object.

source/crc32.c
ilevyor: Why don't we release the memory from the vtable or the rest of the hash struct? is it not worth having an explicit no op function for the purpose of conforming to the model? I guess it's less necessary in c than in an object oriented language.

ilevyor: Does C understand this as a no op function?
static void s_destroy(struct aws_hash *hash);

justinboswell: The vtable is static, it has a permanent home in the data section of the executable, it can't be freed.

justinboswell: > Does C understand this as a no op function?
> static void s_destroy(struct aws_hash *hash);

As long as you put braces after it, yes.

source/crc32.c
justinboswell: Someone else coming in to read this code later will not have just read sha256.c most likely. You want to make it so that someone coming in fresh reads your comment and understands the following code. This would just confuse them :)

source/crc32.c
justinboswell: Can remove this comment

source/crc32.c
```diff
@@ -0,0 +1,99 @@
...
+
+struct aws_hash *aws_crc32_default_new(struct aws_allocator *allocator) {
+    return s_crc32_common_init(allocator, &crc32_vtable);
+}
+
+struct aws_hash *aws_crc32c_default_new(struct aws_allocator *allocator) {
+    return s_crc32_common_init(allocator, &crc32c_vtable);
+}
+
+/* would it be useful to collect most of the logic into a helper function with a flag, and have two versions with the
```
justinboswell, line 56: You could do this and take a function pointer as an argument, just don't do it as a branching (switch/if) implementation, that'll be slower.

justinboswell (approved):

source/crc32.c
```diff
@@ -0,0 +1,95 @@
...
+    uint32_t (*aws_checksums)(const uint8_t *, int, uint32_t)) {
+
+    if (!hash->good) {
+        return aws_raise_error(AWS_ERROR_INVALID_STATE);
+    }
+
+    uintptr_t crc_value = (uintptr_t)hash->impl;
+    uint32_t crc = (uint32_t)crc_value;
+
+    /* hash->impl has type (void *) to match the rest of the API, but we are storing a uintptr, and useing it as such */
```
TingDaoK, line 66: trivial: ~~useing~~ -> using

source/crc32.c
```diff
@@ -0,0 +1,95 @@
...
+    uintptr_t new_crc = aws_checksums(to_hash->ptr, (int)to_hash->len, crc);
```
TingDaoK, line 67: Correct me if I am wrong, the `aws_checksums` returns an uint32_t as a result https://github.com/awslabs/aws-checksums/blob/main/include/aws/checksums/crc.h#L21, but we are treating the result as a pointer here, which seems weird to me.

TingDaoK: Oh, I see the point that we can avoid allocating memory for the result and use the pointer to store it safely. But, at least document it out.

tests/crc_test.c
```diff
@@ -0,0 +1,154 @@
...
+    res |= s_test_known_crc(func_name, func, "DATA_32_ZEROS", &input, KNOWN_CRC32C_32_ZEROES, allocator);
+    res |= s_test_known_crc(func_name, func, "DATA_32_VALUES", &input1, KNOWN_CRC32C_32_VALUES, allocator);
+    res |= s_test_known_crc(func_name, func, "TEST_VECTOR", &input2, KNOWN_CRC32C_TEST_VECTOR, allocator);
+
+    /*this tests three things, first it tests the case where we aren't 8-byte aligned*/
+    /*second, it tests that reads aren't performed before start of buffer*/
+    /*third, it tests that writes aren't performed after the end of the buffer.*/
+    /*if any of those things happen, then the checksum will be wrong and the assertion will fail */
+    uint8_t *s_non_mem_aligned_vector;
+    s_non_mem_aligned_vector = malloc(sizeof(DATA_32_VALUES) + 6);
```
TingDaoK, line 117: I know it's copy/paste from aws-checksum, but still feels like we should avoid use malloc/free and use `aws_mem_acquire`/`aws_mem_release` instead, which is safer. https://github.com/awslabs/aws-c-common/blob/main/include/aws/common/allocator.h#L54-L58

tests/crc_test.c
```diff
@@ -0,0 +1,154 @@
...
+static int s_test_crc32c(struct aws_allocator *allocator, void *ctx) {
+    (void)ctx;
+
+    int res = 0;
+
+    res |= s_test_known_crc32c("aws_crc32c_new", aws_crc32c_new, allocator);
+
+    return res;
+}
+AWS_TEST_CASE(test_crc32c, s_test_crc32c)
```
TingDaoK, line 143: trivial: Maybe rename this test case to something like `crc32c_test` to be more consistent to the reset of tests in this repo?

source/crc32.c
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    size_t buffer_len = output->capacity - output->len;
+
+    if (buffer_len < AWS_CRC32C_LEN) {
+        return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
+    }
+
+    hash->good = false;
+    uintptr_t crc_value = (uintptr_t)hash->impl;
```
graebm, line 93: copy/paste same comment about ptr / int trickery

source/crc32.c
```diff
@@ -0,0 +1,96 @@
...
+    const uint32_t crc = (uint32_t)crc_value;
+    return aws_byte_buf_write(output, (uint8_t *)&crc, sizeof(uint32_t));
```
graebm, line 95: Should we write the result in big-endian to be consistent across machines?

tests/crc_test.c
```diff
@@ -0,0 +1,154 @@
...
+        .ptr = input->ptr,
+    };
+    const struct aws_byte_cursor input_second_half = {
+        .len = (int)(len - len / 2),
+        .ptr = (input->ptr) + (len / 2),
+    };
+    struct aws_hash *crc1 = func(allocator);
+    aws_hash_update(crc1, &input_first_half);
+    aws_hash_update(crc1, &input_second_half);
+    uintptr_t result1 = (uintptr_t)crc1->impl;
```
graebm, line 47: use the public API, even though it's a pain and you know what's stored inside
```
static int s_finalize(struct aws_hash *crc, uint32_t *out_value) {
  uint8_t result_array[AWS_CRC_LEN];
  struct aws_byte_buf buf = aws_byte_init_from_array(result_array, sizeof(result_array));
  ASSERT_SUCCESS(aws_hash_finalize(crc1, &buf, 0));
  struct aws_byte_cursor cursor = aws_cursor_from_byte_buf(&buf);
  if (aws_byte_cursor_read_be32(&cursor, out_value) == false) {
    return aws_raise_error(AWS_ERROR_UNKNOWN);
  }

  return AWS_OP_SUCCESS;
}

```

tests/crc_test.c
```diff
@@ -0,0 +1,154 @@
...
+    for (size_t i = 0; i < len; ++i) {
+        const struct aws_byte_cursor input_i = {
+            .len = 1,
+            .ptr = input->ptr + i,
+        };
+        aws_hash_update(crc2, &input_i);
+    }
+    uintptr_t result2 = (uintptr_t)crc2->impl;
+    ASSERT_HEX_EQUALS(expected, result2, "one byte at a time %s(%s)", func_name, data_name);
+    aws_hash_destroy(crc);
```
graebm, line 60: before you destroy it, call update() and finalize() again and to check that they fail like you expect

`ASSERT_ERROR(AWS_ERROR_INVALID_STATE, aws_hash_update(...))`

tests/crc_test.c
```diff
@@ -0,0 +1,154 @@
...
+    };
+    const struct aws_byte_cursor input1 = {
+        .len = sizeof(DATA_32_VALUES),
+        .ptr = mut_DATA_32_VALUES,
+    };
+    const struct aws_byte_cursor input2 = {
+        .len = sizeof(TEST_VECTOR),
+        .ptr = mut_TEST_VECTOR,
+    };
+    res |= s_test_known_crc(func_name, func, "DATA_32_ZEROS", &input, KNOWN_CRC32C_32_ZEROES, allocator);
```
graebm, line 108: replace  all uses of `res |=`
with
ASSERT_SUCCESS(...);

tests/crc_test.c
```diff
@@ -0,0 +1,154 @@
...
+    s_non_mem_aligned_vector = malloc(sizeof(DATA_32_VALUES) + 6);
+    memset(s_non_mem_aligned_vector, 1, sizeof(DATA_32_VALUES) + 6);
+    memcpy(s_non_mem_aligned_vector + 3, DATA_32_VALUES, sizeof(DATA_32_VALUES));
+
+    const struct aws_byte_cursor input3 = {
+        .len = sizeof(DATA_32_VALUES),
+        .ptr = s_non_mem_aligned_vector + 3,
+    };
+    res |= s_test_known_crc(func_name, func, "non_mem_aligned_vector", &input3, KNOWN_CRC32C_32_VALUES, allocator);
+    free(s_non_mem_aligned_vector);
```
graebm, line 126: aws_mem_release(...)

include/aws/cal/hash.h
```diff
@@ -41,6 +43,15 @@ AWS_CAL_API struct aws_hash *aws_sha256_new(struct aws_allocator *allocator);
  * Allocates and initializes an md5 hash instance.
  */
 AWS_CAL_API struct aws_hash *aws_md5_new(struct aws_allocator *allocator);
+/**
+ * Allocates and initializes a crc32 hash instance
+ */
+AWS_CAL_API struct aws_hash *aws_crc32_new(struct aws_allocator *allocator);
```
graebm, line 49: Add a big old comment explaining HOW users should extract the 32bit number from the buffer passed to finalize. copy/paste comment to the other crc32c_new()

include/aws/cal/hash.h
```diff
@@ -41,6 +43,15 @@ AWS_CAL_API struct aws_hash *aws_sha256_new(struct aws_allocator *allocator);
...
+/**
+ * Allocates and initializes a crc32c hash instance
+ */
+AWS_CAL_API struct aws_hash *aws_crc32c_new(struct aws_allocator *allocator);
+
```
graebm, line 54: you had a great idea. we could add:

```
int aws_crc32_finalize(struct aws_hash *hash, uint32_t *result);
```

cmake/aws-c-cal-config.cmake
```diff
@@ -1,6 +1,7 @@
 include(CMakeFindDependencyMacro)
 
 find_dependency(aws-c-common)
```
JonathanHenson, line 3: aws-checksums already depends on aws-c-common, this line is now redundant

JonathanHenson (commented):
i'm confused about this. Why are we reimplementing aws-checksums in here? Also, a checksum isn't really a hash from a crypto perspective. It was very intentional (requested by crypto team) to not put it in a crypto interface.

graebm:
we're doing this another way

PR #82, opened 2021-04-08
Supports in-source build with AWS-LC.

wps132230:
*Issue:*
Existing in-source build will run CMake scripts via `execute_process` to build and install `aws-lc` during CMake process.
In theory, the configuration step should not build or install anything.
There are several issues need to be solved because of that:
1. CMake requires `sudo` permission if `CMAKE_INSTALL_PREFIX` is not specified, as it installs `aws-lc` into system directory.
2. CMake parameters for `aws-lc.cmake` are very limited, thus not working with some special platforms or cross compile, for example Android build.

*Description of changes:*
Submitted a PR in `aws-lc` to export the target `crypto`: https://github.com/awslabs/aws-lc/pull/127
And updated CMake files here accordingly.


cmake/modules/FindLibCrypto.cmake
```diff
@@ -15,83 +15,93 @@
...
-    ${CMAKE_INSTALL_PREFIX}
-    ${CMAKE_INSTALL_PREFIX}/lib64
-    ${CMAKE_INSTALL_PREFIX}/lib
-    )
+find_package(crypto QUIET)
 
-if (BUILD_SHARED_LIBS)
-    set(LibCrypto_LIBRARY ${LibCrypto_SHARED_LIBRARY})
+if (crypto_FOUND)
+    get_target_property(crypto_INCLUDE_DIR crypto INTERFACE_INCLUDE_DIRECTORIES)
```
justinboswell, line 21: We never export these as LibCrypto_INCLUDE_DIR, etc. Should we so that downstream consumers don't need to worry about 2 targets?

cmake/modules/FindLibCrypto.cmake
```diff
@@ -15,83 +15,93 @@
...
+    message(STATUS "Found target: crypto")
+    message(STATUS "LibCrypto Include Dir: ${crypto_INCLUDE_DIR}")
```
justinboswell, line 23: Should probably replace LibCrypto with crypto, just to differentiate in logs

cmake/modules/FindLibCrypto.cmake
```diff
@@ -15,83 +15,93 @@
...
+find_package(crypto QUIET)
 
-if (BUILD_SHARED_LIBS)
-    set(LibCrypto_LIBRARY ${LibCrypto_SHARED_LIBRARY})
+if (crypto_FOUND)
+    get_target_property(crypto_INCLUDE_DIR crypto INTERFACE_INCLUDE_DIRECTORIES)
+    message(STATUS "aws-c-cal found target: crypto")
+    message(STATUS "crypto Include Dir: ${crypto_INCLUDE_DIR}")
+    set(LIBCRYPTO_FOUND true)
+    set(LibCrypto_FOUND true)
```
sdavtaker, line 25: I would put the comment here of why this is needed. I saw it in another PR, but otherwise I wouldn't know about it.

justinboswell (approved):

cmake/modules/FindLibCrypto.cmake
wps132230: Sure.

PR #54, opened 2020-11-18
Force references to libcrypto functions when linking statically

justinboswell:
prevents dead-stripping

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #89, opened 2021-04-29
Enable runtime use in a musl-based static binary (where dlopen will return null)

elerch:
This is somewhat of a continuation of Issue #54.

These changes remove the assertion that the process module can be read by the dynamic linker. They also add assertions that if the module was not read by the dynamic linker, then symbols for HMAC/MD modules already exist.


tobim:
This patch makes it possible to use the library in static executables. I tested s3 transfers via aws-sdk-cpp and those worked flawlessly. Please consider merging.

source/unix/openssl_platform_init.c
```diff
@@ -122,6 +122,7 @@ bool s_resolve_hmac_102(void *module) {
     if (has_102_symbols) {
         FLOGF("found static libcrypto 1.0.2 HMAC symbols");
     } else {
+        AWS_FATAL_ASSERT(!module && "Cannot read process symbols and libcrypto 1.0.2 HMAC symbols not found");
```
tobim, line 125: These asserts break the dynamically linked library.

jmklix:
We've recently added musl support. Please let us know if you run into any issues when using it

PR #104, opened 2022-01-26
Properly use cmake target exports for libcrypto.

JonathanHenson:


CMakeLists.txt
```diff
@@ -86,25 +86,20 @@ else ()
...
             find_package(OpenSSL REQUIRED)
             find_package(Threads REQUIRED)
-            add_library(crypto UNKNOWN IMPORTED)
-            set_target_properties(crypto PROPERTIES
-                INTERFACE_INCLUDE_DIRECTORIES "${OPENSSL_INCLUDE_DIR}")
-            set_target_properties(crypto PROPERTIES
-                IMPORTED_LINK_INTERFACE_LANGUAGES "C"
-                IMPORTED_LOCATION "${OPENSSL_CRYPTO_LIBRARY}")
-            add_dependencies(crypto Threads::Threads)
+            set(PLATFORM_LIBS OpenSSL::Crypto Threads::Threads)
```
graebm, line 89: why does this branch require "Threads" but the other branches don't?

CMakeLists.txt
```diff
@@ -86,25 +86,20 @@ else ()
...
             message(STATUS "Using libcrypto from system: ${OPENSSL_CRYPTO_LIBRARY}")
         elseif(NOT USE_OPENSSL AND IN_SOURCE_BUILD)
             if (TARGET crypto)
                 message(STATUS "Using libcrypto from AWS-LC")
             else()
                 message(FATAL_ERROR "Target crypto is not defined, failed to find libcrypto.")
             endif()
+            set(PLATFORM_LIBS crypto dl)
```
graebm, line 97: why does this branch require `dl` but the `if (USE_OPENSSL AND NOT ANDROID)` branch doesn't?

I realize the code was like this when you got here, but this series of if-elses is really confusing to me. I see 3 libraries being required-or-not: crypto, thread, dl

And there are 3 toggles that lead to them being selected: USE_OPENSSL, ANDROID, IN_SOURCE_BUILD

could this be written to make more sense?

CMakeLists.txt
```diff
@@ -128,7 +123,9 @@ target_link_libraries(${PROJECT_NAME} PUBLIC ${DEP_AWS_LIBS} ${PLATFORM_LIBS})
 if (BYO_CRYPTO)
     target_compile_definitions(${PROJECT_NAME} PRIVATE -DBYO_CRYPTO)
 elseif (NOT WIN32 AND NOT APPLE)
-    target_include_directories(${PROJECT_NAME} PRIVATE $<TARGET_PROPERTY:crypto,INTERFACE_INCLUDE_DIRECTORIES>)
+    if (TARGET crypto)
+        target_include_directories(${PROJECT_NAME} PRIVATE $<TARGET_PROPERTY:crypto,INTERFACE_INCLUDE_DIRECTORIES>)
```
graebm, line 127: this `target_include_directories()` call would make more sense up in the branch where we call `set(PLATFORM_LIBS crypto dl)` 

rather than trying to reproduce the sequence of if/else/if/else again down here

cmake/modules/FindLibCrypto.cmake
```diff
@@ -15,95 +15,84 @@
...
+    ${CMAKE_INSTALL_PREFIX}/build
+    ${CMAKE_INSTALL_PREFIX}
+    ${CMAKE_INSTALL_PREFIX}/lib64
+    ${CMAKE_INSTALL_PREFIX}/lib
+    )
 
+if (BUILD_SHARED_LIBS)
+    set(LibCrypto_LIBRARY ${LibCrypto_SHARED_LIBRARY})
+else()
+    set(LibCrypto_LIBRARY ${LibCrypto_STATIC_LIBRARY})
```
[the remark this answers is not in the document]
JonathanHenson: This isn't a change in behavior though. Can we do that in a different pass?

CMakeLists.txt
JonathanHenson: attempting to not change behavior, just fix the targets.

CMakeLists.txt
JonathanHenson: once again. Yes. After targets are fixed though?

CMakeLists.txt
JonathanHenson: I might be able to remove these altogether once the targets are fixed. Let me see if I can get everything to build together with it off and I'll just remove all of these.

JonathanHenson: removed

cmake/modules/Findcrypto.cmake
```diff
@@ -0,0 +1,102 @@
...
+# Usage of this module as follows:
+#
+#     find_package(crypto)
+#
+# Variables used by this module, they can change the default behaviour and need
+# to be set before calling find_package:
+#
+# Variables defined by this module:
+#
+#  Crypto_FOUND             System has libcrypto, include and library dirs found
```
graebm, line 12: Let's make the capitalization consistent. Currently it's `crypto` in some places and `Crypto` in others:

Looking at other [cmake find scripts](https://cmake.org/cmake/help/latest/manual/cmake-modules.7.html#find-modules) I see there's not perfect consistency, but `CamelCase` seems most common, followed by `ALLCAPS`. `alllower` is not very common, but if it makes us consistent with other libs we're putting in the `AWS::` namespace then that's fine

but in any case let's pick one style and stick with it

graebm (approved):
fix & ship

cmake/modules/Findcrypto.cmake
JonathanHenson: it's not entirely that simple. A lot of this is due to cmake differences across platforms and case-sensitivity. I'll see if i can get the build working with the pure lower case values, but I might just set both of them to be safe.

PR #112, opened 2022-04-06
Remove AWS:: prefix from crypto module used from cmake path in case of IN_SOURCE_BUILD

serzhick:
*Issue #, if available:*
AWS-SDK-CPP android build failure in case of OpenSSL crypto implementation built before SDK build due to inconsistent linkage against "crypto" and "AWS::crypto" in s2n and aws-c-cal packages:

s2n links against abstract “crypto” thanks to [this workaround](https://github.com/aws/s2n-tls/blame/88c7ae4d3fd9b3e9e49fcecc9bee1ddb8099ae70/CMakeLists.txt#L480-L488)
```
if (TARGET crypto)
    message(STATUS "S2N found target: crypto")
    set(LINK_LIB "crypto")
```

aws-c-common links against "AWS::crypto" (this imported is not related to AWS, btw) thanks to [this hard-coded cmake](https://github.com/awslabs/aws-c-cal/blame/611690ff1087e7e1481480606f3d491f4914f93b/CMakeLists.txt#L101-L103):
```
            set(PLATFORM_LIBS AWS::crypto dl)
```

*Description of changes:*
Make aws-c-cal link against abstract `crypto` just like s2n.

Tested by running AWS-SDK-CPP build for android.


CMakeLists.txt
```diff
@@ -99,8 +99,18 @@ else ()
             set(PLATFORM_LIBS crypto dl)
         else()
             find_package(crypto REQUIRED)
```
JonathanHenson, line 101: if think this should be:

````
if (NOT BYO_CRYPTO)
        file(GLOB AWS_CAL_OS_SRC
            "source/unix/*.c"
        )
        if (USE_OPENSSL AND NOT ANDROID)
            set(PLATFORM_LIBS OpenSSL::Crypto Threads::Threads)
            message(STATUS "Using libcrypto from system: ${OPENSSL_CRYPTO_LIBRARY}")
            if (TARGET crypto)
            else()
                message(FATAL_ERROR "Target crypto is not defined, failed to find libcrypto.")
            endif()
            set(PLATFORM_LIBS crypto dl)
        else()
           #  note aws_use_package() does this for you, except it appends to the public link targets
           # which we probably don't want for this case where we want the crypto dependency private
            if (IN_SOURCE_BUILD)
                set(PLATFORM_LIBS crypto dl)
            else()
                 find_package(crypto REQUIRED)
                 set(PLATFORM_LIBS AWS::crypto dl)
            endif()
        endif()
    endif()
````

JonathanHenson (changes requested):

PR #114, opened 2022-04-08
Find OpenSSL package from system

serzhick:
*Issue #, if available:*
If this library is built as `USE_OPENSSL AND NOT ANDROID`, it's build CMakeLists [configures](https://github.com/awslabs/aws-c-cal/blob/srUseOpenSSL/CMakeLists.txt#L91) to link with `OpenSSL::Crypto`:
```
        if (USE_OPENSSL AND NOT ANDROID)
            set(PLATFORM_LIBS OpenSSL::Crypto Threads::Threads)
```
However, in the cmake script packaged/vended to the user app, it instructs to [only find just `crypto`](https://github.com/awslabs/aws-c-cal/blob/main/cmake/aws-c-cal-config.cmake#L7)

This results in customer app's dependency on OpenSSL::Crypto without OpenSSL being ever found for by cmake.

*Description of changes:*
Reflect on configured built aws-c-cal, if it depends on `OpenSSL::Crypto` then find this one instead of simple `crypto`.

Frankly speaking there might be a better approach, but SDKs/CRTs configuration env variables are lost at this stage (cmake script from installed package is executed completely separately to the SDK/CRT/aws-c-cal make install).

### DIFF
--- a/cmake/aws-c-cal-config.cmake
+++ b/cmake/aws-c-cal-config.cmake
@@ -2,14 +2,21 @@ include(CMakeFindDependencyMacro)
 
 find_dependency(aws-c-common)
 
-if (NOT BYO_CRYPTO AND NOT WIN32 AND NOT APPLE)
-    list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/modules")
-    find_dependency(crypto)
-endif()
-
 if (BUILD_SHARED_LIBS)
     include(${CMAKE_CURRENT_LIST_DIR}/shared/@PROJECT_NAME@-targets.cmake)
 else()
     include(${CMAKE_CURRENT_LIST_DIR}/static/@PROJECT_NAME@-targets.cmake)
 endif()
 
+if (NOT BYO_CRYPTO AND NOT WIN32 AND NOT APPLE)
+    get_target_property(AWS_C_CAL_DEPS AWS::aws-c-cal INTERFACE_LINK_LIBRARIES)
+    # pre-cmake 3.3 IN_LIST search approach
+    list (FIND AWS_C_CAL_DEPS "OpenSSL::Crypto" _index)
+    if (${_index} GREATER -1) # if USE_OPENSSL AND NOT ANDROID
+        find_dependency(OpenSSL REQUIRED)
+        find_dependency(Threads REQUIRED)
+    else()
+        list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/modules")
+        find_dependency(crypto)
+    endif()
+endif()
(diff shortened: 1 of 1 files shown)


cmake/aws-c-cal-config.cmake
```diff
@@ -2,14 +2,21 @@ include(CMakeFindDependencyMacro)
...
-endif()
-
 if (BUILD_SHARED_LIBS)
     include(${CMAKE_CURRENT_LIST_DIR}/shared/@PROJECT_NAME@-targets.cmake)
 else()
     include(${CMAKE_CURRENT_LIST_DIR}/static/@PROJECT_NAME@-targets.cmake)
 endif()
 
+if (NOT BYO_CRYPTO AND NOT WIN32 AND NOT APPLE)
+    get_target_property(AWS_C_CAL_DEPS AWS::aws-c-cal INTERFACE_LINK_LIBRARIES)
```
JonathanHenson, line 12: aws-c-cal target is already entirely resolved by the time this script runs?

cmake/aws-c-cal-config.cmake
```diff
@@ -2,14 +2,21 @@ include(CMakeFindDependencyMacro)
...
+    # pre-cmake 3.3 IN_LIST search approach
+    list (FIND AWS_C_CAL_DEPS "OpenSSL::Crypto" _index)
```
JonathanHenson, line 14: I mean..... by the time we're doing all this..... isn't it simpler to just remove all of the openssl and android branches and see if the problem goes away?

serzhick:
It fixes the build of apps using awa-sdk-cpp in dynamic linking.
Just because this library dynamically links against OpenSSL::Crypto but never calls find of OpenSSL in it's cmake.

cmake/aws-c-cal-config.cmake
serzhick: So basically you suggest to completely drop the support of the build with the OpenSSL.

cmake/aws-c-cal-config.cmake
serzhick: Yes, it is resolved few lines above.

JonathanHenson (approved):

cmake/aws-c-cal-config.cmake
serzhick: I'm really sorry, but it is not "CPP SDK build failure", moreover, SDK itself builds perfectly fine.
It is "apps using aws-c-cal with OpenSSL crypto library form system build failure" because aws-c-cal declares a dependency on OpenSSL but never instructs it's consumers to find it during cmake configure.

serzhick: Comment added.

sdavtaker (approved):

PR #134, opened 2023-02-25
CI job for OpenBSD

knightjoel:
*Issue #, if available:*

N/A

*Description of changes:*

Add a CI job for OpenBSD. This is possible thanks to support being added for OpenBSD to crt-builder. This requires using a custom action to get an OpenBSD instance running (a VM hosted on macOS). Unfortunately, the custom action doesn't do all the necessary things to setup the environment for us so the job has to do some prep work before initiating the build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -146,6 +146,25 @@ jobs:
         chmod a+x builder
         ./builder build -p ${{ env.PACKAGE_NAME }}
 
+  openbsd:
+    runs-on: macos-12 # latest
+    steps:
+      - name: Build ${{ env.PACKAGE_NAME }} + consumers
+        uses: vmactions/openbsd-vm@v0
+        with:
+          mem: 2048
+          prepare: |
+            pkg_add git
+            pkg_add python%3.10
+            pkg_add py3-urllib3
+            pkg_info
+            echo "permit keepenv nopass :wheel as root" > /etc/doas.conf
+            sysctl -n kern.version
+          run: |
+            python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz', 'builder')"
+            chmod a+x builder
+            ./builder build -p ${{ env.PACKAGE_NAME }}
+
   # Test downstream repos.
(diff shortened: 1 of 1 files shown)


knightjoel:
This is a WIP because I'd like to tidy up a couple things in crt-builder first which will allow me to clean up the diff in this PR. As it is, the CI job runs to completion without error, it could just be a little crisper.

.github/workflows/ci.yml
```diff
@@ -146,6 +146,27 @@ jobs:
...
+          architecture: x86-64
+          version: '7.2'
+          shell: bash
+          run: |
+            sudo pkg_add git
+            sudo pkg_add py3-urllib3
+            pkg_info
+            sysctl -n kern.version
+            sudo bash -c "echo \"permit keepenv nopass :wheel as root\" > /etc/doas.conf"
+            python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz', 'builder')"
```
graebm, line 166: can we get aws-crt-builder doing any of this boilerplate?

I guess we're using urllib to acquire the builder... but maybe there's a different shell cmd we can invoke to do the download?

whatever makes it simplest to copy/paste this into other repo's ci.yml files... there's always a certain amount of copy/paste. it's always a balance

knightjoel: Yes! Sort of.

- git: now installed by builder (v0.9.35)
- The pkg_info and sysctl lines: neither are necessary, only for data which might be helpful during troubleshooting. I've removed both lines.
- The `sudo bash` line: Removed as there's no dependency on doas any more (when using builder v0.9.35+).

Using urllib is interesting. While there is at least one command line util in the base OS install we could use to pull builder.pyz down, there's a weird challenge here with OpenBSD's package management tools. There is no mechanism to tell pkg_add "install the latest version of python". You have to give it a version. So it seems the choice is:

1. `pkg_add python%3.10` (for example) and carry the burden forward of bumping this version every time it becomes unavailable in the pkg repo, or
2. Install urllib3 which will pull in a valid version of python as a dependency. We could use the base command line tool to pull down builder.pyz in this case, but someone would likely come along and wonder why urllib is being installed. And I suppose if we continue to use urllib to pull builder.pyz, it looks consistent with the other jobs.

Please give the diff in my latest push a look and let me know what you think.

graebm (approved):

PR #136, opened 2023-03-21
Added 256-bit AES with CBC, CTR, GCM, and Keywrap Modes

JonathanHenson:
Added 256-bit AES with CBC, CTR, GCM, and Keywrap Modes.

This is the final piece required to completely pull the diamond openssl dependency from the C++ SDK, but also provides access to the required algorithms for an encryption client. I'll only be binding out to C++ for now, and it's mostly a port of the existing C++ SymetricCipher abstraction found here:

https://github.com/aws/aws-sdk-cpp/tree/main/src/aws-cpp-sdk-core/source/utils/crypto


source/darwin/commoncrypto_aes.c
```diff
@@ -0,0 +1,701 @@
...
+
+#include <aws/cal/symmetric_cipher.h>
+
+#include <CommonCrypto/CommonCryptor.h>
+#include <CommonCrypto/CommonHMAC.h>
+#include <CommonCrypto/CommonSymmetricKeywrap.h>
+
+#include <aws/cal/private/darwin/common_cryptor_spi.h>
+
+#if defined(__MAC_OS_X_VERSION_MAX_ALLOWED)
```
graebm, line 14: shouldn't we be checking based on ~MAX~ -> `MIN_ALLOWED` ?

I know we compile the CRT sometimes with OSX_DEPLOYMENT_TARGET=10.9

the [cmake docs](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html#variable:CMAKE_OSX_DEPLOYMENT_TARGET) make it seems like this ends up setting a min (no mention of a max)

>  CMake uses this variable value for the -mmacosx-version-min flag

include/aws/cal/symmetric_cipher.h
```diff
@@ -0,0 +1,263 @@
...
+ * If aad is set it will be copied and applied to the cipher.
+ *
+ * Returns NULL on failure. You can check aws_last_error() to get the error code indicating the failure cause.
+ */
+AWS_CAL_API struct aws_symmetric_cipher *aws_aes_gcm_256_new(
+    struct aws_allocator *allocator,
+    const struct aws_byte_cursor *key,
+    const struct aws_byte_cursor *iv,
+    const struct aws_byte_cursor *aad,
+    const struct aws_byte_cursor *decryption_tag);
```
graebm, line 116: This is a lot of arguments. Is there a good chance we'll add more in the future?
if so, consider an options-struct

include/aws/cal/symmetric_cipher.h
```diff
@@ -0,0 +1,263 @@
...
+ * the encrypted output, the call will fail. If you're trying to optimize to use a stack based array
+ * or something, make sure it's at least as large as the size of to_encrypt + an extra BLOCK to account for
+ * padding etc...
+ *
+ * returns AWS_OP_SUCCESS on success. Call aws_last_error() to determine the failure cause if it returns
+ * AWS_OP_ERR;
+ */
+AWS_CAL_API int aws_symmetric_cipher_encrypt(
+    struct aws_symmetric_cipher *cipher,
+    const struct aws_byte_cursor *to_encrypt,
```
graebm, line 150: If a byte-cursor is non-optional pass it by value

```suggestion
   struct aws_byte_cursor to_encrypt,
```

I know we haven't been 100% consistent about this in our code base, but it's better by value. It's obviously non-null, and you can "transform" it like aws_byte_cursor_from_buf() or aws_byte_cursor_from_c_str() without needing a tmp variable

include/aws/cal/symmetric_cipher.h
```diff
@@ -0,0 +1,263 @@
...
+ * the decrypted output, the call will fail. If you're trying to optimize to use a stack based array
+ * or something, make sure it's at least as large as the size of to_decrypt + an extra BLOCK to account for
+ * padding etc...
+ *
+ * returns AWS_OP_SUCCESS on success. Call aws_last_error() to determine the failure cause if it returns
+ * AWS_OP_ERR;
+ */
+AWS_CAL_API int aws_symmetric_cipher_decrypt(
+    struct aws_symmetric_cipher *cipher,
+    const struct aws_byte_cursor *to_decrypt,
```
graebm, line 165: same
```suggestion
   struct aws_byte_cursor to_decrypt,
```

include/aws/cal/symmetric_cipher.h
```diff
@@ -0,0 +1,263 @@
...
+
+/**
+ * Gets the current GMAC tag. If not AES GCM, this function will just copy an empty buffer over.
+ * You would typically call this function after calling aws_symmetric_cipher_finalize_encryption()
+ * to fetch the tag for transmitting it in a cryptographic protocol.
+ *
+ * returns AWS_OP_SUCCESS on success. Call aws_last_error() to determine the failure cause if it returns
+ * AWS_OP_ERR;
+ */
+AWS_CAL_API int aws_symmetric_cipher_get_tag(struct aws_symmetric_cipher *cipher, struct aws_byte_buf *out);
```
graebm, line 216: I'd be fine with this just returning a `struct aws_byte_cursor` by value.
yeah ... it's a bit memory unsafe, but it's so much simpler to use

imagine:
```
JNIEXPORT
jbyteArray JNICALL Java_software_amazon_awssdk_crt_cipher_get_tag(...) {
    ...
    bool success = false;
    struct aws_byte_buf tag_buf;
    aws_byte_buf_init(&tag_buf, aws_jni_get_allocator(), 256/*todo what's a good initial size?*/);
    if (aws_symmetric_cipher_get_tag(cipher)) {
        aws_jni_throw_runtime_exception(env, "failed to get tag");
        goto done;
    }

    success = true;
done:
    jbyteArray result = NULL;
    if (success) {
        struct aws_byte_cursor tag_cursor = aws_byte_cursor_from_buf(&tag_buf);
        result = aws_jni_byte_array_from_cursor(&tag_cursor);
    }
    aws_byte_buf_clean_up(&tag_buf);
    return result;
}
```

vs

```
JNIEXPORT
jbyteArray JNICALL Java_software_amazon_awssdk_crt_cipher_get_tag(...) {
    ...
    struct aws_byte_cursor tag = aws_symmetric_cipher_get_tag(cipher);
    return aws_jni_byte_array_from_cursor(&tag);
}```

**Review by graebm — COMMENTED** [review]
here's what I got as I head off to dinnner

File: include/aws/cal/symmetric_cipher.h [hunk 6]
```diff
@@ -0,0 +1,263 @@
...
+ * Generates a secure random initialization vector of length len_bytes. If is_counter_mode is set, the final 4 bytes
+ * will be reserved as a counter and initialized to 1 in big-endian byte-order.
+ *
+ * out is appended dynamically and will automatically expand the buffer if it needs to. If you need to optimize with
+ * stack allocated arrays or something, make sure it's at least as large as len_bytes.
+ *
+ * returns AWS_OP_SUCCESS on success. Call aws_last_error() to determine the failure cause if it returns
+ * AWS_OP_ERR;
+ */
+AWS_CAL_API int aws_symmetric_cipher_generate_initialization_vector(
```
graebm, line 246: I nitpick this function a lot because it's public
can we make it private?

source/unix/openssl_aes.c
```diff
@@ -0,0 +1,741 @@
...
+            return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
+        }
+    }
+
+    available_write_space = out->capacity - out->len;
+    struct openssl_aes_cipher *openssl_cipher = cipher->impl;
+
+    int len_written = (int)(available_write_space);
+    if (!EVP_EncryptUpdate(
+            openssl_cipher->encryptor_ctx, out->buffer + out->len, &len_written, input->ptr, (int)input->len)) {
```
graebm, line 34: naively casting `input->len` to `int`
- raise error if `len` is too big
- OR run this in a loop, doing the work in chunks

the [docs](https://www.openssl.org/docs/man3.1/man3/EVP_EncryptUpdate.html) say:
> the amount of data written can be anything from zero bytes to (inl + cipher_block_size) bytes

so... I think the max amount of work we can do per call is `INT_MAX - cipher_block_size`???  (or pick something smaller that doesn't push up so hard against the limits)

source/unix/openssl_aes.c
```diff
@@ -0,0 +1,741 @@
...
+    EVP_CIPHER_CTX *encryptor_ctx;
+    EVP_CIPHER_CTX *decryptor_ctx;
+    struct aws_byte_buf working_buffer;
+};
+
+static int s_encrypt(
+    struct aws_symmetric_cipher *cipher,
+    const struct aws_byte_cursor *input,
+    struct aws_byte_buf *out) {
+    size_t required_buffer_space = input->len + cipher->block_size;
```
graebm, line 20: worth doing aws_add_size_checked()?

source/unix/openssl_aes.c
```diff
@@ -0,0 +1,741 @@
...
+    if (available_write_space < required_buffer_space) {
+        if (aws_byte_buf_reserve_relative(out, required_buffer_space) != AWS_OP_SUCCESS) {
+            return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
+        }
+    }
+
+    available_write_space = out->capacity - out->len;
+    struct openssl_aes_cipher *openssl_cipher = cipher->impl;
+
+    int len_written = (int)(available_write_space);
```
graebm, line 32:
```suggestion
    int len_written = 0;
```

it's weird to initialize something to a specific number when it's going to get overwritten. It also confuses me why you picked `required_buffer_space` over `available_write_space`, and also this is a naive cast to int.

so simplest to just set it 0

source/darwin/commoncrypto_aes.c
JonathanHenson: changed the names, but this matches the exact match already done in the C++ SDK. I'll come back to it before merging.

include/aws/cal/symmetric_cipher.h
JonathanHenson: there's not a good chance on these crypto algorithms, and it's a "_new()" function and any version of it needing new arguments will be special enough to warrant another new function.

include/aws/cal/symmetric_cipher.h
JonathanHenson: done

include/aws/cal/symmetric_cipher.h
```diff
@@ -0,0 +1,263 @@
...
+ * Gets the original ke.
+ *
+ * returns AWS_OP_SUCCESS on success. Call aws_last_error() to determine the failure cause if it returns
+ * AWS_OP_ERR;
+ */
+AWS_CAL_API int aws_symmetric_cipher_get_key(struct aws_symmetric_cipher *cipher, struct aws_byte_buf *out);
+
+/**
+ * Generates a secure random initialization vector of length len_bytes. If is_counter_mode is set, the final 4 bytes
+ * will be reserved as a counter and initialized to 1 in big-endian byte-order.
```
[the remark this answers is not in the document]
JonathanHenson: talked offline. I moved this to private. The short answer is CTR and GCM use a 4 byte counter. However, in GCM the 4 byte counter is sort of hidden behind the API because it manages the counter itself. In CTR the counter is exposed in the IV since no hash is associated with it.

source/symmetric_cipher.c
```diff
@@ -0,0 +1,130 @@
...
+void aws_symmetric_cipher_destroy(struct aws_symmetric_cipher *cipher) {
+    cipher->vtable->destroy(cipher);
+}
+
+int aws_symmetric_cipher_encrypt(
+    struct aws_symmetric_cipher *cipher,
+    const struct aws_byte_cursor *to_encrypt,
+    struct aws_byte_buf *out) {
+    if (cipher->good) {
+        return cipher->vtable->encrypt(cipher, to_encrypt, out);
```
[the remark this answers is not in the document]
JonathanHenson: I think I want to leave that decision to the implementations as it isn't part of the overall contract. It represents, the cipher isn't usable, not things like, the input was invalid or something unrelated failed.

source/symmetric_cipher.c
```diff
@@ -0,0 +1,130 @@
...
+
+    /* get the length we wrote back out of it */
+    size_t iv_length = len_bytes;
+
+    if (is_counter_mode) {
+        /* init the counter */
+        /* [nonce 1/4] [ iv 1/2 ] [ctr 1/4] */
+        size_t ctr_start = (iv_length / 2) + (iv_length / 4);
+        out->len = buf_start_len + ctr_start;
+        /* initialize it to a 1, but we don't know the iv size. So just write zeros til the last byte) */
```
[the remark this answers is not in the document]
JonathanHenson: moved it to private. Added an assert to make sure the calculation was defined, and changed the code to just deduct 4 bytes out of the counter.

source/unix/openssl_aes.c
```diff
@@ -0,0 +1,741 @@
...
+
+static int s_encrypt(
+    struct aws_symmetric_cipher *cipher,
+    const struct aws_byte_cursor *input,
+    struct aws_byte_buf *out) {
+    size_t required_buffer_space = input->len + cipher->block_size;
+    size_t available_write_space = out->capacity - out->len;
+
+    if (available_write_space < required_buffer_space) {
+        if (aws_byte_buf_reserve_relative(out, required_buffer_space) != AWS_OP_SUCCESS) {
```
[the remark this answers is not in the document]
JonathanHenson: that's not total desired capacity.

include/aws/cal/symmetric_cipher.h
JonathanHenson: yes and done

source/unix/openssl_aes.c
JonathanHenson: I made it check in the base class before any of these functions are called to check the input. That way the behavior is the same across implementations. Added a unit test as well.

source/unix/openssl_aes.c
JonathanHenson: it's more of an input/output size. For some versions of openssl, this value needs to be the original buffer capacity and then they overwrite it with the amount they wrote. However, i did just refactor all of this so....

source/unix/openssl_aes.c
JonathanHenson: there's a check before this function gets called that guarantees it can't be too large.

source/windows/bcrypt_aes.c
```diff
@@ -0,0 +1,1110 @@
...
+    AWS_ASSERT(s_aes_gcm_algorithm_handle && "BCryptOpenAlgorithmProvider() failed");
+
+    status = BCryptSetProperty(
+        s_aes_gcm_algorithm_handle,
+        BCRYPT_CHAINING_MODE,
+        (PUCHAR)BCRYPT_CHAIN_MODE_GCM,
+        (ULONG)(wcslen(BCRYPT_CHAIN_MODE_GCM) + 1),
+        0);
+
+    AWS_FATAL_ASSERT(NT_SUCCESS(status) && "BCryptSetProperty for GCM chaining mode failed");
```
graebm, line 66: trivial: why the mix of ASSERT() and FATAL_ASSERT() throughout this function?
seems fine to always FATAL_ASSERT, since this function only ever runs once

source/windows/bcrypt_aes.c
```diff
@@ -0,0 +1,1110 @@
...
+
+    BCRYPT_KEY_HANDLE key_handle_to_encrypt =
+        s_import_key_blob(s_aes_keywrap_algorithm_handle, cipher->allocator, &cipher_impl->overflow);
+
+    NTSTATUS status = 0;
+
+    ULONG output_size = 0;
+    /* Call with NULL first to get the required size. */
+    status = BCryptExportKey(
+        key_handle_to_encrypt, cipher_impl->key_handle, BCRYPT_AES_WRAP_KEY_BLOB, NULL, 0, &output_size, 0);
```
graebm, line 950: didn't check whether s_import_key_blob() returned NULL

source/windows/bcrypt_aes.c
```diff
@@ -0,0 +1,1110 @@
...
+    /* If we're in CBC mode the working iv is the same buffer as the iv.
+       so prevent a double free here by checking the addresses first. */
+    bool working_iv_optimized = cipher->iv.buffer == cipher_impl->working_iv.buffer;
+    aws_byte_buf_clean_up_secure(&cipher->key);
+    aws_byte_buf_clean_up_secure(&cipher->iv);
+    aws_byte_buf_clean_up_secure(&cipher->tag);
+    aws_byte_buf_clean_up_secure(&cipher->aad);
+
+    if (!working_iv_optimized) {
+        aws_byte_buf_clean_up_secure(&cipher_impl->working_iv);
```
graebm, line 132: trivial: working_iv has its `allocator` set because it's a by-value copy of `iv`

might be less confusing if the `allocator` wasn't set, when it doesn't actually own its memory

source/windows/bcrypt_aes.c
```diff
@@ -0,0 +1,1110 @@
...
+    struct aws_byte_buf *out) {
+    struct aes_bcrypt_cipher *cipher_impl = cipher->impl;
+
+    if (to_encrypt->len == 0) {
+        return AWS_OP_SUCCESS;
+    }
+
+    size_t predicted_write_length =
+        cipher_impl->cipher_flags & BCRYPT_BLOCK_PADDING
+            ? to_encrypt->len + (AWS_AES_256_CIPHER_BLOCK_SIZE - to_encrypt->len % AWS_AES_256_CIPHER_BLOCK_SIZE)
```
graebm, line 324: utterly trivial: extra parens to make order of operations clear:
```suggestion
            ? to_encrypt->len + (AWS_AES_256_CIPHER_BLOCK_SIZE - (to_encrypt->len % AWS_AES_256_CIPHER_BLOCK_SIZE))
```
right?

source/windows/bcrypt_aes.c
```diff
@@ -0,0 +1,1110 @@
...
+    if (aws_symmetric_cipher_try_ensure_sufficient_buffer_space(out, predicted_write_length)) {
+        return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
+    }
+
+    PUCHAR iv = NULL;
+    ULONG iv_size = 0;
+
+    if (cipher_impl->auth_info_ptr) {
+        iv = cipher_impl->working_iv.buffer;
+        iv_size = (ULONG)cipher_impl->working_iv.capacity;
```
graebm, line 338: trivial: why `capacity` instead of `len`? I get that they're the same at this moment, but we typically use `len`

source/windows/bcrypt_aes.c
```diff
@@ -0,0 +1,1110 @@
...
+    if (to_decrypt->len == 0) {
+        return AWS_OP_SUCCESS;
+    }
+
+    PUCHAR iv = NULL;
+    ULONG iv_size = 0;
+
+    if (cipher_impl->auth_info_ptr) {
+        iv = cipher_impl->working_iv.buffer;
+        iv_size = (ULONG)cipher_impl->working_iv.capacity;
```
graebm, line 446: trivial: same thing about `capacity` vs `len`

source/windows/bcrypt_aes.c
```diff
@@ -0,0 +1,1110 @@
...
+}
+
+static int s_aes_default_encrypt(
+    struct aws_symmetric_cipher *cipher,
+    const struct aws_byte_cursor *to_encrypt,
+    struct aws_byte_buf *out) {
+    struct aes_bcrypt_cipher *cipher_impl = cipher->impl;
+
+    if (to_encrypt->len == 0) {
+        return AWS_OP_SUCCESS;
```
graebm, line 319: the openssl & mac implementations don't check whether input len is 0

should they? or is this a windows quirk?
should we have a unit-test to cover this case?
just asking...

source/windows/bcrypt_aes.c
```diff
@@ -0,0 +1,1110 @@
...
+        NULL,
+        0,
+        cipher_impl->overflow.buffer,
+        (ULONG)cipher_impl->overflow.len,
+        0);
+    (void)status;
+
+    int ret_val = AWS_OP_ERR;
+
+    if (import_key) {
```
graebm, line 1013: if `BCryptImportKey()` fails, and import_key is NULL, we never actually aws_raise_error() before we return AWS_OP_ERR

source/windows/bcrypt_aes.c
```diff
@@ -0,0 +1,1110 @@
...
+            BCRYPT_KEY_DATA_BLOB_HEADER *stream_header = (BCRYPT_KEY_DATA_BLOB_HEADER *)key_data_blob.buffer;
+
+            AWS_FATAL_ASSERT(
+                aws_byte_buf_write(
+                    out, key_data_blob.buffer + sizeof(BCRYPT_KEY_DATA_BLOB_HEADER), stream_header->cbKeyData) &&
+                "Copying key data failed but the allocation should have already occured successfully");
+            ret_val = AWS_OP_SUCCESS;
+
+        } else {
+            cipher->good = false;
```
graebm, line 1047: similar, we never aws_raise_error() when BCryptExportKey() fails

source/unix/openssl_aes.c
```diff
@@ -0,0 +1,723 @@
...
+          EVP_CIPHER_CTX_set_padding(openssl_cipher->encryptor_ctx, 0)) ||
+        !(EVP_DecryptInit_ex(openssl_cipher->decryptor_ctx, EVP_aes_256_gcm(), NULL, NULL, NULL) &&
+          EVP_DecryptInit_ex(
+              openssl_cipher->decryptor_ctx,
+              NULL,
+              NULL,
+              openssl_cipher->cipher_base.key.buffer,
+              openssl_cipher->cipher_base.iv.buffer) &&
+          EVP_CIPHER_CTX_set_padding(openssl_cipher->decryptor_ctx, 0))) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
```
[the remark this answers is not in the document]
JonathanHenson: The only thing that will make these calls fail is invalid arguments and the crypto lib doesn't give better errors than "invalid argument"

source/darwin/commoncrypto_aes.c
```diff
@@ -0,0 +1,674 @@
...
+        cc_cipher->cipher_base.key.len,
+        NULL,
+        0,
+        0,
+        kCCModeOptionCTR_BE,
+        &cc_cipher->encryptor_handle);
+
+#ifdef MAC_10_13_AVAILABLE
+    status |=
+        CCCryptorGCMSetIV(cc_cipher->encryptor_handle, cc_cipher->cipher_base.iv.buffer, cc_cipher->cipher_base.iv.len);
```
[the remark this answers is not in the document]
JonathanHenson: It will just fail, but it's plenty simple to move it to separate branches.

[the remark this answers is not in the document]
JonathanHenson: fixed

include/aws/cal/private/symmetric_cipher_priv.h
```diff
@@ -0,0 +1,65 @@
...
+    void *impl;
+};
+
+AWS_EXTERN_C_BEGIN
+
+/**
+ * Generates a secure random initialization vector of length len_bytes. If is_counter_mode is set, the final 4 bytes
+ * will be reserved as a counter and initialized to 1 in big-endian byte-order.
+ *
+ * returns AWS_OP_SUCCESS on success. Call aws_last_error() to determine the failure cause if it returns
```
DmitriyMusatkin, line 44: nit: doc does not seem to match function as return is void

source/cal.c
```diff
@@ -33,7 +33,10 @@ static struct aws_error_info s_errors[] = {
     AWS_DEFINE_ERROR_INFO_CAL(
         AWS_ERROR_CAL_UNSUPPORTED_ALGORITHM,
         "The specified algorithim is unsupported on this platform."),
-};
+    AWS_DEFINE_ERROR_INFO_CAL(
```
DmitriyMusatkin, line 36: should we add error type for tag/aad not matching in GCM?

source/windows/bcrypt_aes.c
```diff
@@ -0,0 +1,1121 @@
...
+    size_t iv_size,
+    bool is_ctr_mode,
+    bool is_gcm) {
+
+    if (!cipher->cipher.key.len) {
+        if (key) {
+            aws_byte_buf_init_copy_from_cursor(&cipher->cipher.key, cipher->cipher.allocator, *key);
+        } else {
+            aws_byte_buf_init(&cipher->cipher.key, cipher->cipher.allocator, AWS_AES_256_KEY_BYTE_LEN);
+            aws_symmetric_cipher_generate_key(AWS_AES_256_KEY_BYTE_LEN, &cipher->cipher.key);
```
DmitriyMusatkin, line 170: should we pull AWS_AES_256_KEY_BYTE_LEN into args or at least state in doc that this function assumes 256 byte keys. could save some pain in the future with someone using it for different key length

source/unix/openssl_aes.c
```diff
@@ -0,0 +1,717 @@
...
+    }
+
+    return ret_val;
+}
+
+static int s_init_gcm_cipher_materials(struct aws_symmetric_cipher *cipher) {
+    struct openssl_aes_cipher *openssl_cipher = cipher->impl;
+
+    if (!(EVP_EncryptInit_ex(openssl_cipher->encryptor_ctx, EVP_aes_256_gcm(), NULL, NULL, NULL) &&
+          EVP_EncryptInit_ex(
```
DmitriyMusatkin, line 350: is there something specific about gcm that requires splitting init into 2 calls? i dont see openssl docs mentioning the need for it

DmitriyMusatkin (approved):

include/aws/cal/symmetric_cipher.h
```diff
@@ -0,0 +1,216 @@
...
+
+/**
+ * Gets the original intialization vector as a cursor.
+ * The memory in this cursor is unsafe as it refers to the internal buffer.
+ * This was done because the use case doesn't require fetching these during an
+ * encryption or decryption operation and it dramatically simplifies the API.
+ *
+ * Unlike some other fields, this value does not change after the inital construction of the cipher.
+ */
+AWS_CAL_API struct aws_byte_cursor aws_symmetric_cipher_get_initialization_vector(
```
TingDaoK, line 194: Can this data be unavailable, eg: `aws_aes_keywrap_256_new` doesn't take iv? Are we just return empty cursor? If so, can we add document for that?

TingDaoK (approved):

source/cal.c
JonathanHenson: I don't know how to detect a bad tag is the reason the encryption operations failed, and IIRC cryptographers don't tell you that due to various tampering attack vectors being able to use that information to craft a better attack.

source/unix/openssl_aes.c
JonathanHenson: yes, it's openssl API across multiple versions protection. This is the code that we've been able to work across all versions.

source/cal.c
DmitriyMusatkin: According to docs if EVP_DecryptFinal fails in AEAD mode then it can be assumed authn part failed and im guessing mac/win would have something similar. Havent worked with GCM at c level, but i know that java/ts exposed bad tag as exception. But if thats no longer a recommended practice, im fine with hiding it. 

relevant part from openssl docs:
When decrypting, the return value of EVP_DecryptFinal() or EVP_CipherFinal() indicates whether the operation was successful. If it does not indicate success, the authentication operation has failed and any output data MUST NOT be used as it is corrupted.
https://www.openssl.org/docs/man3.0/man3/EVP_DecryptFinal_ex.html

PR #145, opened 2023-04-20
Fix for osx deployment target 10.9

waahm7:
*Issue #, if available:*

*Description of changes:*
Python until version 3.9 sets `CMAKE_OSX_DEPLOYMENT_TARGET` to `10.9`, so we must continue to support it for now. This pull request fixes some warnings that arise when `-DCMAKE_OSX_DEPLOYMENT_TARGET=10.9` is added.

The `API_AVAILABLE` macro doesn't work because it is equivalent to using the `@available` attribute on a declaration. `@available` is equivalent to `__builtin_available` in C which is just a runtime check and can not be used this way.
source: https://developer.apple.com/documentation/swift/marking-api-availability-in-objective-c , https://epir.at/2019/10/30/api-availability-and-target-conditionals/, https://developer.apple.com/forums/thread/719208

Move the `darwin` specific header to the source directory because `Swift` compiles everything in the include directory by default. Currently, there is no good way to exclude headers.


JonathanHenson:
I just undid this two days ago because it was incorrect. Could you please verify and add an explanation about what was incorrect?

https://github.com/awslabs/aws-c-cal/pull/140

source/darwin/commoncrypto_aes.c
```diff
@@ -355,6 +353,37 @@ struct aws_symmetric_cipher *aws_aes_ctr_256_new_impl(
     return &cc_cipher->cipher_base;
 }
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
JonathanHenson, line 357: if your macro above is correct, you should never be warned of deprecated declarations.

source/darwin/commoncrypto_aes.c
```diff
@@ -355,6 +353,37 @@ struct aws_symmetric_cipher *aws_aes_ctr_256_new_impl(
...
+/*
+ * Note that CCCryptorGCMFinal is deprecated in Mac 10.13. It also doesn't compare the tag with expected tag
+ * https://opensource.apple.com/source/CommonCrypto/CommonCrypto-60118.1.1/include/CommonCryptorSPI.h.auto.html
+ */
+static CCStatus s_cc_crypto_gcm_finalize(struct _CCCryptor *encryptor_handle, uint8_t *buffer, size_t tag_length) {
+#ifdef USE_LATEST_CRYPTO_API
+    if (__builtin_available(macOS 10.13, iOS 11.0, *)) {
```
JonathanHenson, line 364: this makes sense, my only hold-back is the initial versioning macro at the top

source/darwin/commoncrypto_aes.c
```diff
@@ -355,6 +353,45 @@ struct aws_symmetric_cipher *aws_aes_ctr_256_new_impl(
...
+
+static CCCryptorStatus s_cc_cryptor_gcm_set_iv(struct _CCCryptor *encryptor_handle, uint8_t *buffer, size_t length) {
+#ifdef USE_LATEST_CRYPTO_API
+    if (__builtin_available(macOS 10.13, iOS 11.0, *)) {
+        return CCCryptorGCMSetIV(encryptor_handle, buffer, length);
+    } else {
+/* We would never hit this branch for newer macOS and iOS versions because of the __builtin_available check, so we can
+ * suppress the compiler warning. */
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
graebm, line 386: holy crap this all is ugly
but I can't come up with anything better
nice work

graebm (approved):

JonathanHenson (approved):

source/darwin/commoncrypto_aes.c
waahm7: Discussed offline, we need this for the runtime check.

source/darwin/commoncrypto_aes.c
waahm7: Discussed offline, and added more details in the PR description.

PR #146, opened 2023-04-20
Added validation for key and iv sizes.

JonathanHenson:


SergeyRyabinin (approved):

source/symmetric_cipher.c
```diff
@@ -85,17 +85,42 @@ static bool s_check_input_size_limits(const struct aws_symmetric_cipher *cipher,
...
+    const struct aws_byte_cursor *iv,
+    size_t expected_iv_size) {
+    if (key && key->len != expected_key_size) {
+        aws_raise_error(AWS_ERROR_CAL_INVALID_KEY_LENGTH_FOR_ALGORITHM);
+        return false;
+    }
+
+    if (iv && iv->len != expected_iv_size) {
+        aws_raise_error(AWS_ERROR_CAL_INVALID_CIPHER_MATERIAL_SIZE_FOR_ALGORITHM);
+        return false;
```
graebm, line 100: trivial: let's be consistent with how we usually report errors, and have this function return `AWS_OP_SUCCESS/ERR` instead of `bool`

mixing and matching error-handling styles is confusing

graebm (approved):
fix & shipit

source/symmetric_cipher.c
JonathanHenson: yeah, i thought that, and then tried the alternative. It's worse. Either way the check has to be done. and i didn't like doing it with int return type.

JonathanHenson: And that's consistent with the function 3 lines above it.

JonathanHenson: it's not like we can just return the result. it just doesn't make sense and it's less ergonomic and it's 100% internal.

PR #148, opened 2023-05-01
Support BoringSSL

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
support building aws-c-cal against boringssl

### DIFF
--- a/include/aws/cal/private/opensslcrypto_common.h
+++ b/include/aws/cal/private/opensslcrypto_common.h
@@ -14,6 +14,8 @@ typedef void (*hmac_ctx_clean_up)(HMAC_CTX *);
 typedef int (*hmac_ctx_update)(HMAC_CTX *, const unsigned char *, size_t);
 typedef int (*hmac_ctx_final)(HMAC_CTX *, unsigned char *, unsigned int *);
 
+typedef void (*crypto_generic_fn_ptr)(void);
+
 struct openssl_hmac_ctx_table {
     hmac_ctx_new new_fn;
     hmac_ctx_free free_fn;
@@ -25,8 +27,8 @@ struct openssl_hmac_ctx_table {
     hmac_ctx_reset reset_fn;
 
     struct {
-        void (*reset_fn)(void);
-        void (*init_ex_fn)(void);
+        crypto_generic_fn_ptr reset_fn;
+        crypto_generic_fn_ptr init_ex_fn;
     } impl;
 };
 
--- a/include/aws/cal/private/opensslcrypto_common.h
+++ b/include/aws/cal/private/opensslcrypto_common.h
@@ -30,9 +30,9 @@ struct openssl_hmac_ctx_table {
     hmac_ctx_reset reset_fn;
 
     /* There is slight variance between the crypto interfaces.
-        Note that function pointer casting is undefined behavior. 
+        Note that function pointer casting is undefined behavior.
         To workaround the issue, use generic pointer for crypto and let delegate
-        function cast it back to correct type. 
+        function cast it back to correct type.
         Do not use following fields manually. */
     struct {
         crypto_generic_fn_ptr reset_fn;
--- a/include/aws/cal/private/opensslcrypto_common.h
+++ b/include/aws/cal/private/opensslcrypto_common.h
@@ -23,6 +23,11 @@ struct openssl_hmac_ctx_table {
     hmac_ctx_update update_fn;
     hmac_ctx_final final_fn;
     hmac_ctx_reset reset_fn;
+
+    struct {
+        void (*)() reset_fn;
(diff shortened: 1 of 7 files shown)


source/unix/openssl_platform_init.c
```diff
@@ -48,6 +69,25 @@ extern int HMAC_Update(HMAC_CTX *, const unsigned char *, size_t) __attribute__(
...
+    AWS_PRECONDITION(ctx);
+
+    int (*reset_ptr)(HMAC_CTX *) = (int (*)(HMAC_CTX *))g_aws_openssl_hmac_ctx_table->impl.reset_fn;
+    reset_ptr(ctx);
+}
+
+static int s_hmac_init_ex_openssl(HMAC_CTX *ctx, const void *key, size_t key_len, const EVP_MD *md, ENGINE *impl) {
+    AWS_PRECONDITION(ctx);
+
+    int (*init_ex_ptr)(HMAC_CTX *, const void *, int, const EVP_MD *, ENGINE *) =
```
graebm, line 82: add comment like 
`/* openssl takes key_len as an int, rather than size_t */`

also, need to defend against `key_len > INT_MAX`

source/unix/openssl_platform_init.c
```diff
@@ -48,6 +69,25 @@ extern int HMAC_Update(HMAC_CTX *, const unsigned char *, size_t) __attribute__(
 extern int HMAC_Final(HMAC_CTX *, unsigned char *, unsigned int *) __attribute__((weak, used));
 extern int HMAC_Init_ex(HMAC_CTX *, const void *, int, const EVP_MD *, ENGINE *) __attribute__((weak, used));
 
+static void s_hmac_ctx_reset_openssl(HMAC_CTX *ctx) {
+    AWS_PRECONDITION(ctx);
+
+    int (*reset_ptr)(HMAC_CTX *) = (int (*)(HMAC_CTX *))g_aws_openssl_hmac_ctx_table->impl.reset_fn;
+    reset_ptr(ctx);
```
graebm, line 76: add comment like
`/* openssl returns int, but other implementations don't */`

also, we should probably AWS_FATAL_ASSERT that it returns 0

include/aws/cal/private/opensslcrypto_common.h
```diff
@@ -23,6 +28,16 @@ struct openssl_hmac_ctx_table {
...
     hmac_ctx_final final_fn;
     hmac_ctx_reset reset_fn;
+
+    /* There is slight variance between the crypto interfaces.
+        Note that function pointer casting is undefined behavior.
+        To workaround the issue, use generic pointer for crypto and let delegate
+        function cast it back to correct type.
+        Do not use following fields manually. */
+    struct {
+        crypto_generic_fn_ptr reset_fn;
```
graebm, line 38: I don't actually see the reset_fn() ever getting called from anywhere? Are we actually using it? If not, can we remove all the reset_fn related code?

source/unix/openssl_platform_init.c
graebm: Yikes, looks like the only place that calls this function naively casts it to an int, we should change that

https://github.com/awslabs/aws-c-cal/blob/55e478b31cf50517cd359c1ef2e79ef6c2a1f9d0/source/unix/opensslcrypto_hmac.c#L76

include/aws/cal/private/opensslcrypto_common.h
DmitriyMusatkin: you are right, looks like it was used at some point, but has since been removed. removing it alltogether

source/unix/openssl_platform_init.c
```diff
@@ -48,6 +60,23 @@ extern int HMAC_Update(HMAC_CTX *, const unsigned char *, size_t) __attribute__(
 extern int HMAC_Final(HMAC_CTX *, unsigned char *, unsigned int *) __attribute__((weak, used));
 extern int HMAC_Init_ex(HMAC_CTX *, const void *, int, const EVP_MD *, ENGINE *) __attribute__((weak, used));
 
+static int s_hmac_init_ex_openssl(HMAC_CTX *ctx, const void *key, size_t key_len, const EVP_MD *md, ENGINE *impl) {
+    AWS_PRECONDITION(ctx);
+    if (key_len > INT_MAX) {
+        return 1;
```
graebm, line 66:
```suggestion
        return 0;
```
https://www.openssl.org/docs/man3.0/man3/HMAC_Init_ex.html
> HMAC_CTX_reset(), HMAC_Init_ex(), HMAC_Update(), HMAC_Final() and HMAC_CTX_copy() return 1 for success or 0 if an error occurred.

graebm (approved):
fix & ship

PR #160, opened 2023-09-19
Be more explicit on int format in DER

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Fix 0 handling. 
Add functions to explicitly indicate what int format is expected


source/der.c
```diff
@@ -36,11 +36,19 @@ struct der_tlv {
     uint8_t *value;
 };
 
-static void s_decode_tlv(struct der_tlv *tlv) {
+static int s_decode_tlv(struct der_tlv *tlv) {
     if (tlv->tag == AWS_DER_INTEGER) {
         uint8_t first_byte = tlv->value[0];
-        /* if the first byte is 0, it just denotes unsigned and should be removed */
-        if (first_byte == 0x00) {
+        if (first_byte & 0x80) {
```
graebm, line 42: don't we need to check that `length > 0` before checking `value[0]`?

graebm (approved):
fix & ship

source/der.c
```diff
@@ -98,6 +102,10 @@ static int s_der_read_tlv(struct aws_byte_cursor *cur, struct der_tlv *tlv) {
         len = len_bytes;
     }
 
+    if (len > cur->len) {
+        return aws_raise_error(AWS_ERROR_CAL_MALFORMED_ASN1_ENCOUNTERED);
+    }
```
graebm, line 107: Is it OK that this function incremented the cursor?
Are there any assumptions in the calling code that the cursor is unchanged if the function fails?

If this error results in the calling code just erroring out, then shipit.

If the calling code tries to do something with the cursor after an error, then NO-shipit

graebm (approved):
conditional shipit

source/der.c
DmitriyMusatkin: yeah, this function is invoked when parsing der during decoder_new. any error in decoding just gets propagated back up and we return NULL decoder back. failing to read length field also results in that flow.

PR #163, opened 2023-10-11
Use fake pub key

TingDaoK:
We use SecCreateKeyWithData to create ECC key. Expected format for the key passed to that api is a byte buffer consisting of "0x04 | x | y | p", where x,y is public pair and p is private key. 

Note that in some cases we only have private key (as we derived it as described above) and we need to construct SecKey from that. 

In those cases our code just passes 0,0 point for x,y, i.e. "0x04 | 0 | 0 | p". 

This used to work on Macs before 14, but in 14+ SecCreateKeyWithData returns error, which is reasonable since 0,0 is not a valid public point.

To get around the issue, we use a fake public key, which is a valid public point, but not matching the private key as a quick workaround.


source/darwin/securityframework_ecc.c
```diff
@@ -207,14 +230,46 @@ struct aws_ecc_key_pair *aws_ecc_key_pair_new_from_private_key_impl(
...
+     * consisting of "0x04 | x | y | p", where x,y is public pair and p is private key.
+     *
+     * In this case we only have private key and we need to construct SecKey from that.
+     *
+     * We used to just pass 0,0 point for x,y, i.e. "0x04 | 0 | 0 | p".
+     *
+     * This used to work on Macs before 14, but in 14+ SecCreateKeyWithData returns error,
+     * which is reasonable since 0,0 is not a valid public point.
+     *
+     * To get around the issue, we use a fake public key, which is a valid public point, but not matching the private
```
graebm, line 244: 💯 great comment block

source/darwin/securityframework_ecc.c
```diff
@@ -207,14 +230,46 @@ struct aws_ecc_key_pair *aws_ecc_key_pair_new_from_private_key_impl(
...
+            fake_pub_x = aws_byte_cursor_from_array(FAKE_X_ECDSA_P256, sizeof(FAKE_X_ECDSA_P256));
+            fake_pub_y = aws_byte_cursor_from_array(FAKE_Y_ECDSA_P256, sizeof(FAKE_Y_ECDSA_P256));
+            break;
+        case AWS_CAL_ECDSA_P384:
+            fake_pub_x = aws_byte_cursor_from_array(FAKE_X_ECDSA_P384, sizeof(FAKE_X_ECDSA_P384));
+            fake_pub_y = aws_byte_cursor_from_array(FAKE_Y_ECDSA_P384, sizeof(FAKE_Y_ECDSA_P384));
+            break;
+
+        default:
+            break;
```
graebm, line 262: raise error?

source/darwin/securityframework_ecc.c
```diff
@@ -21,6 +21,29 @@ static uint8_t s_preamble = 0x04;
 
 static size_t s_der_overhead = 8;
 
+/* The hard-coded "valid" public keys. */
+static uint8_t FAKE_X_ECDSA_P256[] = {
```
JonathanHenson, line 25: Is this the correct naming convention? I thought it was s_lower_snake_case

source/darwin/securityframework_ecc.c
```diff
@@ -207,14 +230,46 @@ struct aws_ecc_key_pair *aws_ecc_key_pair_new_from_private_key_impl(
...
+            break;
+    }
     struct commoncrypto_ecc_key_pair *cc_key_pair =
-        s_alloc_pair_and_init_buffers(allocator, curve_name, empty_cur, empty_cur, *priv_key);
+        s_alloc_pair_and_init_buffers(allocator, curve_name, fake_pub_x, fake_pub_y, *priv_key);
 
     if (!cc_key_pair) {
         return NULL;
     }
+    /* Clear out the public keys in the key pair */
```
graebm, line 270:
```suggestion
    /* Clear out the fake public keys in the key pair */
```

graebm (approved):
fix & ship

source/darwin/securityframework_ecc.c
```diff
@@ -21,6 +21,29 @@ static uint8_t s_preamble = 0x04;
 
 static size_t s_der_overhead = 8;
 
+/* The hard-coded "valid" public keys. */
```
DmitriyMusatkin, line 24: nit: add a note that its the same public keys we use for tests. dont want anyone getting confused about us checking in real keys into source code

source/darwin/securityframework_ecc.c
TingDaoK: lol, copy/pasted from @DmitriyMusatkin 's write up in the ticket. 👍

source/darwin/securityframework_ecc.c
```diff
@@ -207,14 +230,46 @@ struct aws_ecc_key_pair *aws_ecc_key_pair_new_from_private_key_impl(
...
+     * To get around the issue, we use a fake public key, which is a valid public point, but not matching the private
+     * key as a quick workaround.
+     */
+    struct aws_byte_cursor fake_pub_x;
+    AWS_ZERO_STRUCT(fake_pub_x);
+    struct aws_byte_cursor fake_pub_y;
+    AWS_ZERO_STRUCT(fake_pub_y);
+    switch (curve_name) {
+        case AWS_CAL_ECDSA_P256:
+            fake_pub_x = aws_byte_cursor_from_array(FAKE_X_ECDSA_P256, sizeof(FAKE_X_ECDSA_P256));
```
DmitriyMusatkin, line 253: nit: AWS_ARRAY_SIZE is probably a slightly better choice than sizeof

source/darwin/securityframework_ecc.c
DmitriyMusatkin: we can fatal assert here? curve being supported should already have been verified

TingDaoK: I thought about that, but seems like our API doesn't really raise error for the unknown enum. But, yeah, I can raise the error here.

DmitriyMusatkin (approved):

source/darwin/securityframework_ecc.c
```diff
@@ -207,14 +230,46 @@ struct aws_ecc_key_pair *aws_ecc_key_pair_new_from_private_key_impl(
...
+    }
     struct commoncrypto_ecc_key_pair *cc_key_pair =
-        s_alloc_pair_and_init_buffers(allocator, curve_name, empty_cur, empty_cur, *priv_key);
+        s_alloc_pair_and_init_buffers(allocator, curve_name, fake_pub_x, fake_pub_y, *priv_key);
 
     if (!cc_key_pair) {
         return NULL;
     }
+    /* Clear out the public keys in the key pair */
+    aws_byte_buf_clean_up(&cc_key_pair->key_pair.pub_x);
```
DmitriyMusatkin, line 271: nit: if you are going to clean up, do a secure one, even though it does not particularly matter in this case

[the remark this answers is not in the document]
TingDaoK: Updated to zero out after we use the buf to create SecKey

TingDaoK: updated to zero out instead

PR #164, opened 2023-10-12
Fix build against openssl 1.0.2

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
in 1.0.2 evperr header did not existing and errors were instead defined in evp.h

### DIFF
--- a/source/unix/openssl_rsa.c
+++ b/source/unix/openssl_rsa.c
@@ -11,7 +11,10 @@
 #include <openssl/evp.h>
 
 #if !defined(OPENSSL_IS_AWSLC) && !defined(OPENSSL_IS_BORINGSSL)
+/*Error defines were part of evp.h in 1.0.x and were moved to evperr.h in 1.1.0*/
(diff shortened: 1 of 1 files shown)


TingDaoK (approved):

source/unix/openssl_rsa.c
```diff
@@ -11,7 +11,10 @@
 #include <openssl/evp.h>
 
 #if !defined(OPENSSL_IS_AWSLC) && !defined(OPENSSL_IS_BORINGSSL)
```
waahm7, line 13: nit: I think we should probably use defined(OPENSSL_IS_OPENSSL) instead. This check may break in the future if we add more variants.

waahm7 (approved):

source/unix/openssl_rsa.c
DmitriyMusatkin: openssl does not define OPENSSL_IS_OPENSSL. we do define it manually in another place to the exact same thing as above.

PR #183, opened 2024-05-06
Fix handling of empty plaintext for AES

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*


source/unix/openssl_aes.c
```diff
@@ -14,8 +14,21 @@ struct openssl_aes_cipher {
...
 static int s_encrypt(struct aws_symmetric_cipher *cipher, struct aws_byte_cursor input, struct aws_byte_buf *out) {
 
+    /*
+     * Openssl 1.1.1 and its derivatives like aws-lc and boringssl do not handle
+     * the case of null input of 0 len gracefully in update (it succeeds, but
+     * finalize after it will fail). Openssl 3.0 fixed this. Other crypto implementations
+     * do not have similar issue.
+     * To workaround the issue, replace null cursor with empty cursor.
+     */
+    if (input.ptr == NULL && input.len == 0) {
```
waahm7, line 28: Don't know the context much, but should we handle the case where the length is 0 and ptr is pointed to garbage? If yes, then
```suggestion
    if (input.len == 0) {
```

waahm7 (approved):

source/unix/openssl_aes.c
DmitriyMusatkin: its kinda grey area. i though we didnt consider cursor with garbage ptr and 0 len a valid cursor. but looks like anything with 0 len is valid. it does not seem to matter much as openssl can handle anything non-null, but sure for simplicity we can remove null check from this if

PR #186, opened 2024-05-22
Implement runtime check on libcrypto linkage

WillChilds-Klein:
# Description

This PR adds a runtime check to ensure that the preprocess/compile-time libcrypto is of the same "family" as linked libcrypto. With this check, we enforce that aws-c-cal is compiled against supported libcrypto's (unless `BYO_CRYPTO` has been specified). In the case of compilation against AWS-LC, we also validate that compile/link version is the same. We do not check version in the OpenSSL case as discussed [here](https://github.com/awslabs/aws-c-cal/pull/186#discussion_r1612497699) and [here](https://github.com/awslabs/aws-c-cal/pull/186#discussion_r1630300266).

# Testing

This PR's CI jobs verify the "happy" path for supported libcryptos (AWS-LC, BoringSSL, BYO_CRYPTO, and a few versions of OpenSSL). I've also tested the failure case with aws-c-cal compiled against OpenSSL 1.1.1, then linked by aws-crt-python's build to AWS-LC. Run from aws-crt-python's source root (with debug logging configured in aws-c-cal):

```
$ python3 -m unittest discover --failfast --verbose
...
Fatal error condition occurred in /tmp/tmp.5xrtKAz4xw/tmp/src/Aws-c-cal/source/unix/openssl_platform_init.c:591: strcmp(expected_version, runtime_version) == 0 && "libcrypto mislink"
Exiting Application
[DEBUG] [2024-05-23T23:53:49Z] [00007f8dc4ba1740] [libcrypto_resolve] - Compiled with libcrypto OpenSSL 1.1.1x  30 Jan 2024, linked to libcrypto AWS-LC FIPS 1.26.0
...
```

---


source/unix/openssl_platform_init.c
```diff
@@ -555,6 +559,32 @@ static enum aws_libcrypto_version s_resolve_libcrypto_lib(void) {
...
 }
 
+static void s_validate_libcrypto_linkage(void) {
+    /* NOTE: the choice of stack buffer size is somewhat arbitrary. it's
+     * possible, but unlikely, that libcrypto version strings may exceed this in
+     * the future. we guard against buffer overflow by limiting write size in
+     * snprintf with the size of the buffer itself. if libcrypto version strings
+     * do eventually exceed the chosen size, this runtime check will fail and
+     * will need to be addressed by increasing buffer size.*/
+    char expected_version[32] = { 0 };
```
graebm, line 569: trivial: Why cut it so close? The sample string on the [OpenSSL docs](https://www.openssl.org/docs/man1.0.2/man3/SSLeay_version.html) "OpenSSL 1.0.1a 15 Oct 2015" was 26 chars.
```suggestion
    char expected_version[64] = { 0 };
```

source/unix/openssl_platform_init.c
```diff
@@ -555,6 +559,32 @@ static enum aws_libcrypto_version s_resolve_libcrypto_lib(void) {
...
+#elif defined(OPENSSL_IS_BORINGSSL)
+    snprintf(expected_version, sizeof(expected_version), "BoringSSL");
+#elif defined(OPENSSL_IS_OPENSSL)
+    snprintf(expected_version, sizeof(expected_version), OPENSSL_VERSION_TEXT);
+#else
+#error Unsupported libcrypto!
+#endif
+    const char *runtime_version = SSLeay_version(SSLEAY_VERSION);
+    AWS_FATAL_ASSERT(strlen(expected_version) == strlen(runtime_version) && "libcrypto mislink");
+    AWS_FATAL_ASSERT(strstr(expected_version, runtime_version) && "libcrypto mislink");
```
graebm, line 585: is there a particular reason not to simply use `strcmp()`?
```suggestion
    AWS_FATAL_ASSERT(strcmp(expected_version, runtime_version) == 0 && "libcrypto mislink");
```

source/unix/openssl_platform_init.c
```diff
@@ -555,6 +559,32 @@ static enum aws_libcrypto_version s_resolve_libcrypto_lib(void) {
     return AWS_LIBCRYPTO_NONE;
 }
 
+static void s_validate_libcrypto_linkage(void) {
```
graebm, line 562:
```suggestion
/* Validate at runtime that we're linked against the same libcrypto we compiled against. */
static void s_validate_libcrypto_linkage(void) {
```

source/unix/openssl_platform_init.c
WillChilds-Klein: No, agreed `strcmp` makes more sense.

source/unix/openssl_platform_init.c
```diff
@@ -555,6 +559,37 @@ static enum aws_libcrypto_version s_resolve_libcrypto_lib(void) {
...
+#elif !defined(BYO_CRYPTO)
+#    error Unsupported libcrypto!
+#endif
+    const char *runtime_version = SSLeay_version(SSLEAY_VERSION);
+    AWS_LOGF_DEBUG(
+        AWS_LS_CAL_LIBCRYPTO_RESOLVE,
+        "Compiled with libcrypto %s, linked to libcrypto %s",
+        expected_version,
+        runtime_version);
+    AWS_FATAL_ASSERT(strcmp(expected_version, runtime_version) == 0 && "libcrypto mislink");
```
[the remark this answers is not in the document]
WillChilds-Klein: To better illustrate this, I've implemented the above suggestion [here](https://github.com/awslabs/aws-c-cal/pull/186/commits/8cbf6d752ddb569db526eeb78b48c83f2e108f33) in this PR. Let me know what the team thinks, we can always revert to the stricter validation if that's preferred.

source/unix/openssl_platform_init.c
```diff
@@ -583,6 +634,8 @@ static enum aws_libcrypto_version s_resolve_libcrypto(void) {
         result = s_resolve_libcrypto_lib();
     }
 
+    s_validate_libcrypto_linkage();
```
WillChilds-Klein, line 637: Another call-out for reviewers -- if we're concerned about this affecting downstream consumers that have have cross-linked libcryptos, we can consider making this check an opt-in CMake flag. Perhaps something like `-DVALIDATE_LIBCRYPTO_LINK_VERSION=ON`.

source/unix/openssl_platform_init.c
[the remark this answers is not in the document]
WillChilds-Klein: Discussed offline with @DmitriyMusatkin. We may want to relax this check to only validate the libcrypto "family" matches between compile and link, and allow flexibility for library versions as OpenSSL has relatively stable ABI across minor (and major?) versions.

source/unix/openssl_platform_init.c
```diff
@@ -555,6 +559,53 @@ static enum aws_libcrypto_version s_resolve_libcrypto_lib(void) {
...
+static void s_validate_libcrypto_linkage(void) {
+    /* NOTE: the choice of stack buffer size is somewhat arbitrary. it's
+     * possible, but unlikely, that libcrypto version strings may exceed this in
+     * the future. we guard against buffer overflow by limiting write size in
+     * snprintf with the size of the buffer itself. if libcrypto version strings
+     * do eventually exceed the chosen size, this runtime check will fail and
+     * will need to be addressed by increasing buffer size.*/
+    char expected_version[64] = {0};
+#if defined(OPENSSL_IS_AWSLC)
+    /* get FIPS mode at runtime becuase headers don't give any indication of
```
graebm, line 572: utterly trivial:
```suggestion
    /* get FIPS mode at runtime because headers don't give any indication of
```

source/unix/openssl_platform_init.c
graebm: Fair call-out, but for simplicity's sake let's have it on by default

source/unix/openssl_platform_init.c
```diff
@@ -555,6 +559,53 @@ static enum aws_libcrypto_version s_resolve_libcrypto_lib(void) {
...
+     * and minor version.*/
+    AWS_FATAL_ASSERT(strstr(runtime_version, "OpenSSL") == &runtime_version[0] && "runtime libcrypto not OpenSSL");
+    const unsigned long expected_openssl_version = OPENSSL_VERSION_NUMBER;
+    const unsigned long expected_openssl_major = (expected_openssl_version & 0x00000000F0000000L);
+    const unsigned long expected_openssl_minor = (expected_openssl_version & 0x000000000FF00000L);
+    const unsigned long runtime_openssl_version = OpenSSL_version_num();
+    const unsigned long runtime_openssl_major = (runtime_openssl_version & 0x00000000F0000000L);
+    const unsigned long runtime_openssl_minor = (runtime_openssl_version & 0x000000000FF00000L);
+    AWS_FATAL_ASSERT(expected_openssl_major == runtime_openssl_major && "OpenSSL major version mismatch");
+    AWS_FATAL_ASSERT(expected_openssl_minor == runtime_openssl_minor && "OpenSSL minor version mismatch");
```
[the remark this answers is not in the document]
DmitriyMusatkin: thing to note is that we try to workaround that abi breakage in openssl_init.c, but trying to detect at runtime which version we have and then dlsym the correct api signature. 
because of that i think we openssl we can have relaxed checking that just matches the family. otherwise we might break people who mix and match compile and runtime versions.
in future, we can consider having a strict minor and major version check and removing that dlsym code altogether

source/unix/openssl_platform_init.c
```diff
@@ -555,6 +559,47 @@ static enum aws_libcrypto_version s_resolve_libcrypto_lib(void) {
...
+        expected_version,
+        runtime_version);
+#if defined(OPENSSL_IS_OPENSSL)
+    /* Validate that the string "AWS-LC" doesn't appear in OpenSSL version str. */
+    AWS_FATAL_ASSERT(strstr("AWS-LC", expected_version) == NULL);
+    AWS_FATAL_ASSERT(strstr("AWS-LC", runtime_version) == NULL);
+    /* Validate both expected and runtime versions begin with OpenSSL's version str prefix. */
+    const char *openssl_prefix = "OpenSSL ";
+    AWS_FATAL_ASSERT(strncmp(openssl_prefix, expected_version) == 0);
+    AWS_FATAL_ASSERT(strncmp(openssl_prefix, runtime_version) == 0);
```
graebm, line 597: strncmp takes 3 arguments

source/unix/openssl_platform_init.c
```diff
@@ -555,6 +559,47 @@ static enum aws_libcrypto_version s_resolve_libcrypto_lib(void) {
...
+        "Compiled with libcrypto %s, linked to libcrypto %s",
+        expected_version,
+        runtime_version);
+#if defined(OPENSSL_IS_OPENSSL)
+    /* Validate that the string "AWS-LC" doesn't appear in OpenSSL version str. */
+    AWS_FATAL_ASSERT(strstr("AWS-LC", expected_version) == NULL);
+    AWS_FATAL_ASSERT(strstr("AWS-LC", runtime_version) == NULL);
+    /* Validate both expected and runtime versions begin with OpenSSL's version str prefix. */
+    const char *openssl_prefix = "OpenSSL ";
+    AWS_FATAL_ASSERT(strncmp(openssl_prefix, expected_version) == 0);
```
graebm, line 596: debatable: why are we asserting on the `expected_version`? This seems like overkill, we just set this like 10 lines ago

WillChilds-Klein: i put the `AWS_FATAL_ASSERT(strstr(...) == NULL);` asserts in there because  AWS-LC's [`OPENSSL_VERSION_TEXT`](https://github.com/aws/aws-lc/blob/401329930889cca10fa34bc066512bce091e1b2c/include/openssl/crypto.h#L116) sort of "looks like" OpenSSL's and I wanted to ensure that we're neither compiling nor linking against AWS-LC when we think we're compiling/linking against OpenSSL.

the `AWS_FATAL_ASSERT(strncmp(...) == 0);` asserts are the first actual check that we are indeed compiling/linking against OpenSSL (or at least something that looks like it; i.e. has a `OpenSSL ` version string prefix).

these assertions should be cheap, and are only run once on startup.

graebm (approved):

DmitriyMusatkin (approved):

PR #197, opened 2024-08-08
Use prebuilt aws-lc

sfod:
*Issue #, if available:*

Downstream projects can now build and install aws-lc at configuration time. aws-c-cal needs to support this explicitly.
See https://github.com/awslabs/aws-crt-cpp/pull/648

*Description of changes:*

Detect and use prebuilt aws-lc.


CMakeLists.txt
```diff
@@ -101,10 +101,14 @@ else ()
         elseif(NOT USE_OPENSSL AND IN_SOURCE_BUILD)
             if (TARGET crypto)
                 message(STATUS "Using libcrypto from AWS-LC")
+                set(PLATFORM_LIBS crypto)
+            elseif(aws-lc-PREBUILT)
```
bretambrose, line 105: I would use a different symbol name here.  Mixing the casing and using `-` feels weird even if cmake is ok with it.

bretambrose (approved):

PR #206, opened 2025-02-12
Ed25519 support.

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
support ed25519 on linux through libcrypto and experimental support for mac/win through libcrypto


include/aws/cal/ed25519.h
```diff
@@ -0,0 +1,86 @@
...
+     * This will only export the key block, framing (i.e. pem) is left as exercise for the caller.
+     * b64 encoding is done as convenience since common framing formats require it.
+     */
+    AWS_CAL_ED25519_KEY_EXPORT_OPENSSH_B64,
+};
+
+/*
+ * Get public key for the key pair.
+ * Key in specified format is appended to the buffer.
+ * use aws_ed25519_key_pair_get_public_key_size to figure out how much capacity buffer needs for a given format.
```
graebm, line 55:
```suggestion
 * The buffer must be initialized before this call, with sufficient capacity to hold the result.
 * Use aws_ed25519_key_pair_get_public_key_size() to figure out how much capacity buffer needs for a given format.
```

include/aws/cal/ed25519.h
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Gets the size of the exported public key.
+ */
+AWS_CAL_API size_t aws_ed25519_key_pair_get_public_key_size(enum aws_ed25519_key_export_format format);
+
+/*
+ * Get private key for the key pair.
+ * Key in specified format is appended to the buffer.
+ * use aws_ed25519_key_pair_get_private_key_size to figure out how much capacity buffer needs for a given format.
```
graebm, line 70:
```suggestion
 * The buffer must be initialized before this call, with sufficient capacity to hold the result.
 * Use aws_ed25519_key_pair_get_private_key_size() to figure out how much capacity buffer needs for a given format.
```

source/shared/ed25519-lc.c
```diff
@@ -0,0 +1,351 @@
...
+#if defined(OPENSSL_IS_OPENSSL) && OPENSSL_VERSION_NUMBER <= 0x10101000L
+    /* ed25519 support does not exist prior to 1.1.1 */
+    aws_raise_error(AWS_ERROR_CAL_UNSUPPORTED_ALGORITHM);
+    return NULL;
+#else
+    EVP_PKEY *pkey = NULL;
+
+    EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_id(EVP_PKEY_ED25519, NULL);
+    if (ctx == NULL) {
+        aws_raise_error(AWS_ERROR_CAL_CRYPTO_OPERATION_FAILED);
```
graebm, line 46: for all these OpenSSL functions, it would be nice to log specifics of what failed, similar to https://github.com/awslabs/aws-c-cal/blob/7299c6ab9244595b140d604475cdd6c6921be8ae/source/unix/openssl_rsa.c#L55

source/shared/ed25519-lc.c
```diff
@@ -0,0 +1,351 @@
...
+
+on_error:
+    EVP_PKEY_CTX_free(ctx);
+    aws_raise_error(AWS_ERROR_CAL_CRYPTO_OPERATION_FAILED);
+    return NULL;
+#endif
+}
+
+struct aws_ed25519_key_pair *aws_ed25519_key_pair_acquire(struct aws_ed25519_key_pair *key_pair) {
+    return aws_ref_count_acquire(&key_pair->ref_count);
```
graebm, line 74:
```suggestion
    if (key_pair != NULL) {
        aws_ref_count_acquire(&key_pair->ref_count);
    }
    return key_pair;
```

source/shared/ed25519-lc.c
```diff
@@ -0,0 +1,351 @@
...
+        aws_ref_count_release(&key_pair->ref_count);
+    }
+    return NULL;
+}
+
+static struct aws_byte_cursor s_key_type_literal = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("ssh-ed25519");
+
+int s_ed25519_openssh_encode_public_key(const struct aws_ed25519_key_pair *key_pair, struct aws_byte_buf *out) {
+    if (!aws_byte_buf_write_be32(out, s_key_type_literal.len) ||
+        aws_byte_buf_append(out, &s_key_type_literal) != AWS_OP_SUCCESS || !aws_byte_buf_write_be32(out, 32)) {
```
graebm, line 88: If a function fails, the length of the buffer it writes to should remain unchanged.

But this function can fail midway.

Either have this goto error and put the length back where it was, or have the public functions that use these helpers do that if any of the helpers fail

source/shared/ed25519-lc.c
```diff
@@ -0,0 +1,351 @@
...
+        return aws_raise_error(AWS_ERROR_CAL_CRYPTO_OPERATION_FAILED);
+    }
```
graebm, line 90: should this be AWS_ERROR_SHORT_BUFFER?

source/shared/ed25519-lc.c
```diff
@@ -0,0 +1,351 @@
...
+
+    size_t pub_len = 32;
+    AWS_FATAL_ASSERT(out->capacity - out->len >= pub_len);
```
graebm, line 93: properly raise error and return,
(or have a length check earlier that makes it OK to simply assert here

source/shared/ed25519-lc.c
```diff
@@ -0,0 +1,351 @@
...
+    if (aws_base64_encode(&key_cur, out) != AWS_OP_SUCCESS) {
+        return AWS_OP_ERR;
+    }
+
+    return AWS_OP_SUCCESS;
+}
+
+int s_ed25519_export_public_raw(const struct aws_ed25519_key_pair *key_pair, struct aws_byte_buf *out) {
+    size_t remaining = out->capacity - out->len;
+    if (remaining < 32) {
```
graebm, line 129: debatable: should we be more flexible and call EVP_PKEY_get_raw_public_key() with NULL to learn the size?

source/shared/ed25519-lc.c
```diff
@@ -0,0 +1,351 @@
...
+    switch (format) {
+        case AWS_CAL_ED25519_KEY_EXPORT_RAW:
+            return s_ed25519_export_public_raw(key_pair, out);
+        case AWS_CAL_ED25519_KEY_EXPORT_OPENSSH_B64:
+            return s_ed25519_export_public_openssh(key_pair, out);
+        default:
+            return aws_raise_error(AWS_ERROR_CAL_UNSUPPORTED_KEY_FORMAT);
+    }
+
+    return AWS_OP_SUCCESS;
```
graebm, line 158: trivial: not necessary, since switch has a `default:`
```suggestion
```

source/shared/ed25519-lc.c
```diff
@@ -0,0 +1,351 @@
...
+        aws_byte_buf_clean_up(&key_buf);
+        return AWS_OP_ERR;
+    }
+
+    aws_byte_buf_clean_up(&key_buf);
+    return AWS_OP_SUCCESS;
+
+on_error:
+    aws_byte_buf_clean_up(&key_buf);
+    return aws_raise_error(AWS_ERROR_CAL_CRYPTO_OPERATION_FAILED);
```
graebm, line 303: it would be nice to raise specific errors: SHORT_BUFFER vs AWS_ERROR_RANDOM_GEN_FAILED vs AWS_ERROR_CAL_CRYPTO_OPERATION_FAILED, instead of this blanket one

source/shared/ed25519-lc.c
DmitriyMusatkin: im not the biggest fan of calling it twice. we know the size must be 32 and we have buffer preallocated for that, if it wrote less bytes than that, we know something went wrong

source/shared/ed25519-lc.c
DmitriyMusatkin: changed a bunch of places where we can definitely say why it failed. the rest are generic crypto op failed.

source/shared/ed25519-lc.c
```diff
@@ -0,0 +1,379 @@
...
+
+struct aws_ed25519_key_pair *aws_ed25519_key_pair_new_generate(struct aws_allocator *allocator) {
+#if defined(OPENSSL_IS_OPENSSL) && OPENSSL_VERSION_NUMBER <= 0x10101000L
+    /* ed25519 support does not exist prior to 1.1.1 */
+    aws_raise_error(AWS_ERROR_CAL_UNSUPPORTED_ALGORITHM);
+    return NULL;
+#else
+    EVP_PKEY *pkey = NULL;
+
+    AWS_LOGF_DEBUG(0, "here1");
```
TingDaoK, line 44: debug log

source/shared/ed25519-lc.c
```diff
@@ -0,0 +1,379 @@
...
+
+    EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_id(EVP_PKEY_ED25519, NULL);
+    if (ctx == NULL) {
+        aws_raise_error(AWS_ERROR_CAL_CRYPTO_OPERATION_FAILED);
+        return NULL;
```
TingDaoK, line 49: maybe
`aws_reinterpret_lc_evp_error_as_crt(0//error, "EVP_PKEY_CTX_new_id", AWS_LS_CAL_ED25519)`?

source/shared/ed25519-lc.c
```diff
@@ -0,0 +1,379 @@
...
+    aws_ref_count_init(&key_pair->ref_count, key_pair, s_ed25519_destroy_key);
+    key_pair->allocator = allocator;
+    key_pair->key = pkey;
+
+    EVP_PKEY_CTX_free(ctx);
+    return key_pair;
+
+on_error:
+    EVP_PKEY_CTX_free(ctx);
+    aws_raise_error(AWS_ERROR_CAL_CRYPTO_OPERATION_FAILED);
```
TingDaoK, line 71: we already raised in `aws_reinterpret_lc_evp_error_as_crt`?

source/shared/ed25519-lc.c
```diff
@@ -0,0 +1,379 @@
...
+
+    aws_mem_release(lc_key_pair->allocator, lc_key_pair);
+}
+
+struct aws_ed25519_key_pair *aws_ed25519_key_pair_new_generate(struct aws_allocator *allocator) {
+#if defined(OPENSSL_IS_OPENSSL) && OPENSSL_VERSION_NUMBER <= 0x10101000L
+    /* ed25519 support does not exist prior to 1.1.1 */
+    aws_raise_error(AWS_ERROR_CAL_UNSUPPORTED_ALGORITHM);
+    return NULL;
+#else
```
TingDaoK, line 41: Trivial: why not just end if here?

source/shared/ed25519-lc.c
```diff
@@ -0,0 +1,379 @@
...
+
+int s_ed25519_openssh_encode_public_key(const struct aws_ed25519_key_pair *key_pair, struct aws_byte_buf *out) {
+    if (!aws_byte_buf_write_be32(out, (uint32_t)s_key_type_literal.len) ||
+        aws_byte_buf_append(out, &s_key_type_literal) != AWS_OP_SUCCESS) {
+        return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
+    }
+
+    if (!aws_byte_buf_write_be32(out, 32) ||
+        aws_ed25519_key_pair_get_public_key(key_pair, AWS_CAL_ED25519_KEY_EXPORT_RAW, out) != AWS_OP_SUCCESS) {
+        return aws_raise_error(AWS_ERROR_CAL_CRYPTO_OPERATION_FAILED);
```
TingDaoK, line 100: `aws_byte_buf_write_be32` doesn't raise error, but the other function does. Maybe separate them, so that we don't override the error that already raised?

source/shared/ed25519-lc.c
DmitriyMusatkin: EVP_PKEY_CTX_new_id does not set any error params as far as i can tell, so running it through reinterp is not useful

source/shared/ed25519-lc.c
DmitriyMusatkin: win flagged the rest of function as unreachable

source/shared/ed25519-lc.c
DmitriyMusatkin: i was trying to minimize the amount of if statements, but it was not supper successful. refactored it a bit to make more clear

source/shared/ed25519-noop.c
```diff
@@ -0,0 +1,54 @@
...
+    (void)key_pair;
+    (void)format;
+    (void)out;
+    return aws_raise_error(AWS_ERROR_PLATFORM_NOT_SUPPORTED);
+}
+
+size_t aws_ed25519_key_pair_get_private_key_size(enum aws_ed25519_key_export_format format) {
+    (void)format;
+    AWS_FATAL_ASSERT(0);
+    return 0;
```
waahm7: Other no-op functions raise an error but these two functions just always use a fatal assert to fail. Any reason why?

source/shared/ed25519-lc.c
waahm7: Pulling these into a constants might be helpful since we are using it multiple places.

source/shared/ed25519-lc.c
```diff
@@ -0,0 +1,374 @@
...
+}
+
+size_t aws_ed25519_key_pair_get_public_key_size(enum aws_ed25519_key_export_format format) {
+    switch (format) {
+        case AWS_CAL_ED25519_KEY_EXPORT_RAW:
+            return 32;
+        case AWS_CAL_ED25519_KEY_EXPORT_OPENSSH_B64:
+            return 68;
+        default:
+            return 0;
```
waahm7:
```suggestion
        default:
            aws_raise_error(AWS_ERROR_CAL_UNSUPPORTED_KEY_FORMAT);
            return 0;
```

tests/ed25519_test.c
```diff
@@ -0,0 +1,69 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#include <aws/cal/ed25519.h>
+
+#include <aws/testing/aws_test_harness.h>
+
+#include "test_case_helper.h"
+
+static int s_ed25519_key_pair_generate_test(struct aws_allocator *allocator, void *ctx) {
```
waahm7, line 11: Is there a way to verify the generated values instead of just the length?

waahm7 (approved):

source/shared/ed25519-noop.c
DmitriyMusatkin: no-op should always return null keys, so if you are calling those its likely you are doing something wrong. doing fatal here was mikes suggestion above

tests/ed25519_test.c
DmitriyMusatkin: not in an easy way. its either just reconstructing the format and ignoring parts that change or mocking lc pieces to generate the same keys.

Issue #213, opened 2025-04-10
`dlopen(NULL, RTLD_NOW)` fails on statically linked executables with musl


kyusic:
### Describe the bug

When using `aws-c-cal` on a statically linked executable built with musl (e.g., using the `-static` flag), running the resulting binary sometimes results in a fatal error with the following message:

```
process && "Unable to load symbols from process space"
```

After investigation, the issue appears to originate from the following code:

https://github.com/awslabs/aws-c-cal/blob/856c828621d85ce98cc35a912b1167774c2f652e/source/unix/openssl_platform_init.c#L643-L647

In this scenario, `dlopen(NULL, RTLD_NOW)` returns `NULL`, which triggers the `AWS_FATAL_ASSERT` and causes the program to abort.

It seems that musl libc returns `NULL` when `dlopen(NULL, ...)` is called from a statically linked executable. This doesn't occur with glibc. As a result, `aws-c-cal` cannot currently be used to build statically linked executables on Alpine Linux, which uses musl.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

It would be helpful if `aws-c-cal` could provide an alternative mechanism when running in statically linked environments on musl-based systems.

### Current Behavior

```
Fatal error condition occurred in /root/aws-c-cal/source/unix/openssl_platform_init.c:647: process && "Unable to load symbols from process space"
Exiting Application
No call stack information available
Aborted (core dumped)
```

### Reproduction Steps

On Alpine linux 3.21 x86_64,

```shellsession
$ apk add g++ git cmake make openssl-libs-static
$ git clone https://github.com/awslabs/aws-c-common.git
$ cmake -S aws-c-common -B aws-c-common/build -DCMAKE_INSTALL_PREFIX=$HOME/opt/aws
$ cmake --build aws-c-common/build --target install
$ cmake -S aws-c-cal -B aws-c-cal/build -DCMAKE_INSTALL_PREFIX=$HOME/opt/aws -DCMAKE_PREFIX_PATH=$HOME/opt/aws
$ cmake --build aws-c-cal/build --target install
$ cat >test.c <<EOF
#include <aws/cal/cal.h>

int main()
{
	aws_cal_library_init(aws_aligned_allocator());	
}
EOF
$ gcc -static test.c -I $HOME/opt/aws/include -L $HOME/opt/aws/lib -laws-c-cal -laws-c-common -lcrypto
$ ./a.out
[...]

PR #215, opened 2025-04-15
dlopen(NULL) returns NULL on static linked executable

TingDaoK:
*Issue #, if available:*

- https://github.com/awslabs/aws-c-cal/issues/213
- CI reproduces this https://github.com/awslabs/aws-c-cal/actions/runs/14479191863/job/40612167525?pr=216

*Description of changes:*

- When compile into static, `dlopen(NULL)` gonna return `NULL`, however, GLIBC seems only generates warnings, but still give the handle to the current process back, so this issue was not found by us
- The current logic to load libcrypto symbols doesn't need to be this complicated. If libcrypto is linked/loaded, we don't need to load current process to get the symbol, we can just look up the weak reference symbol directly
- Only when the symbols cannot be found, we can try to searching for shared lib and loading that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,7 +136,7 @@ jobs:
   linux-openssl3-static:
     runs-on: ubuntu-24.04 # latest
     steps:
-    - uses: aws-actions/configure-aws-credentials@v4  
+    - uses: aws-actions/configure-aws-credentials@v4
       with:
         role-to-assume: ${{ env.CRT_CI_ROLE }}
         aws-region: ${{ env.AWS_DEFAULT_REGION }}
@@ -149,7 +149,7 @@ jobs:
   linux-openssl-shared:
     runs-on: ubuntu-24.04 # latest
     steps:
-    - uses: aws-actions/configure-aws-credentials@v4  
+    - uses: aws-actions/configure-aws-credentials@v4
       with:
         role-to-assume: ${{ env.CRT_CI_ROLE }}
         aws-region: ${{ env.AWS_DEFAULT_REGION }}
@@ -171,7 +171,7 @@ jobs:
   linux-aws-lc-fips:
     runs-on: ubuntu-24.04 # latest
     steps:
-    - uses: aws-actions/configure-aws-credentials@v4  
+    - uses: aws-actions/configure-aws-credentials@v4
       with:
         role-to-assume: ${{ env.CRT_CI_ROLE }}
         aws-region: ${{ env.AWS_DEFAULT_REGION }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -171,20 +171,57 @@ jobs:
...
+          # x64 builds
+          - architecture: x64
+            image: alpine-3.16-x64
+            static: false
+          - architecture: x64
+            image: alpine-3.16-x64
+            static: true
+          # arm builds
+          - architecture: arm
+            image: alpine-3.16-armv7
```
DmitriyMusatkin, line 198: does arm v7 actually run faster on ubuntu arm?

source/unix/openssl_platform_init.c
```diff
@@ -38,6 +38,7 @@ static struct aws_allocator *s_libcrypto_allocator = NULL;
  * and avoid dead-stripping
  */
 #if defined(OPENSSL_IS_AWSLC) || defined(OPENSSL_IS_BORINGSSL)
+/* TODO: MUSL LTO will strip the symbols even with the weak refs */
```
DmitriyMusatkin, line 41: can you expand on this comment with what we found?
i.e. that on alpine used weak ref does not seem to be enough and lto (or maybe some other optimization) still strips those symbols if it thinks they are not used

source/unix/openssl_platform_init.c
```diff
@@ -140,8 +141,12 @@ bool s_resolve_hmac_102(void *module) {
     /* were symbols bound by static linking? */
     bool has_102_symbols = init_fn && clean_up_fn && update_fn && final_fn && init_ex_fn;
     if (has_102_symbols) {
-        AWS_LOGF_DEBUG(AWS_LS_CAL_LIBCRYPTO_RESOLVE, "found static libcrypto 1.0.2 HMAC symbols");
+        AWS_LOGF_DEBUG(AWS_LS_CAL_LIBCRYPTO_RESOLVE, "found loaded libcrypto 1.0.2 HMAC symbols");
```
DmitriyMusatkin, line 144: found loaded symbols sounds kinda weird. maybe found "weak ref" symbols?
and then somewhere before then we should say that we are trying to resolve in process symbols

DmitriyMusatkin (approved):

.github/workflows/ci.yml
TingDaoK: that was copy/paste, I'll remove it.

However, it seems like weirdly working in the CI. I didn't have qemu enabled, but it runs armv7 docker image from armv8 host.

PR #223, opened 2025-08-15
SHA512 support

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -310,8 +310,8 @@ jobs:
   macos-ed25519:
     strategy:
       matrix:
-        image: [macos-14-large, macos-14]
-    name: ${{ matrix.image == 'macos-14' && 'macos' || 'macos-x64' }} with lc ed25519
+        image: [macos-15-large, macos-15]
+    name: ${{ matrix.image == 'macos-15' && 'macos' || 'macos-x64' }} with lc ed25519
     runs-on: ${{ matrix.image }}
     steps:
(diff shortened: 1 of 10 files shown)


source/windows/bcrypt_hash.c
```diff
@@ -86,17 +108,38 @@ static void s_load_md5_alg_handle(void *user_data) {
...
+    bcrypt_hash->hash.impl = bcrypt_hash;
+    bcrypt_hash->hash.digest_size = AWS_SHA512_LEN;
+    bcrypt_hash->hash.good = true;
+    bcrypt_hash->hash_obj = hash_obj;
+    NTSTATUS status = BCryptCreateHash(
+        s_sha512_alg, &bcrypt_hash->hash_handle, bcrypt_hash->hash_obj, (ULONG)s_sha512_obj_len, NULL, 0, 0);
+
+    if (((NTSTATUS)status) < 0) {
+        aws_mem_release(allocator, bcrypt_hash);
         return NULL;
```
TingDaoK, line 130: https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptcreatehash
This function seems to have some error code, if we can convert them to our error code and then return null, it will be more consistent with the other code base.

But, it seems to be like this for c-call all the time. Maybe we can add a todo or something for better error handling in c-cal.

TingDaoK (approved):

PR #229, opened 2025-10-13
Refactor ec key import

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Replace the logic of iterating through the asn1 structure looking for fields that seem to match expected ec key, with approach of looking for several concrete key structures.

### DIFF
--- a/source/ecc.c
+++ b/source/ecc.c
@@ -576,11 +576,11 @@ int aws_der_decoder_load_ecc_key_pair(
     struct aws_byte_cursor *out_private_d,
     enum aws_ecc_curve_name *out_curve_name) {
 
-    AWS_PRECONDITION(decoder);
-    AWS_PRECONDITION(out_public_x_coord);
-    AWS_PRECONDITION(out_public_y_coord);
-    AWS_PRECONDITION(out_private_d);
-    AWS_PRECONDITION(out_curve_name);
+    AWS_ERROR_PRECONDITION(decoder);
+    AWS_ERROR_PRECONDITION(out_public_x_coord);
+    AWS_ERROR_PRECONDITION(out_public_y_coord);
+    AWS_ERROR_PRECONDITION(out_private_d);
+    AWS_ERROR_PRECONDITION(out_curve_name);
 
     /**
      * Since this is a generic api to parse from ans1, we can encounter several key structures.


source/ecc.c
```diff
@@ -248,22 +236,78 @@ static int s_der_decoder_load_ecc_private_key_pair_from_sec1(
...
+
+            if (aws_der_decoder_tlv_type(decoder) == AWS_DER_OBJECT_IDENTIFIER) {
+                if (aws_der_decoder_tlv_blob(decoder, &oid)) {
+                    return aws_raise_error(AWS_ERROR_CAL_MALFORMED_ASN1_ENCOUNTERED);
+                }
+                if (aws_ecc_curve_name_from_oid(&oid, &curve_name)) {
+                    return aws_raise_error(AWS_ERROR_CAL_UNKNOWN_OBJECT_IDENTIFIER);
+                }
+                *curve_name_set = true;
+                aws_der_decoder_next(decoder); /* skip to field after */
```
TingDaoK: ditto

source/der.c
```diff
@@ -400,6 +400,15 @@ struct aws_der_decoder *aws_der_decoder_new(struct aws_allocator *allocator, str
     return NULL;
 }
 
+struct aws_der_decoder *aws_der_decoder_nested_tlv_decoder(struct aws_der_decoder *decoder) {
+    struct aws_byte_cursor cursor;
+    AWS_ZERO_STRUCT(cursor);
+    if (aws_der_decoder_tlv_string(decoder, &cursor)) {
```
TingDaoK: I am not familiar with the der decoder, but looks like this is only used once, and seems like a very specific case the nested der is a string type?

source/unix/opensslcrypto_ecc.c
```diff
@@ -228,15 +228,55 @@ struct aws_ecc_key_pair *aws_ecc_key_pair_new_generate_random(
...
+    }
+
+    EC_POINT_free(point);
+    BN_free(pub_x_num);
+    BN_free(pub_y_num);
+
+    return AWS_OP_SUCCESS;
+
+on_error:
+    EC_POINT_free(point);
```
TingDaoK: Loos like our previous logic has a null protection, but not here. 
According to Google AI summary:
```
Safe to use: You can pass NULL to BN_free() without any risk. It is a safe operation, [says the Ubuntu Manpage](https://manpages.ubuntu.com/manpages/noble/man3/BN_clear_free.3ssl.html).
```
Wow, the doc it linked seems legit.

source/der.c
DmitriyMusatkin: Yeah, this is a bit of unusual case. 
basically instead of inlining the fields of the nested structure, it is turned into octet string.
we need to parse the nested structure, so we need to read the string and start a new decoder. 
there are legacy reasons for why its done this way, but they are kinda boring. 

this is a helper to initialize a new decoder from a string at current location. we need it once now. there are more keys in the wild that have similar structure, so it might be helpful in general.
im not tied to this, it was just simpler without having to expose allocator for decoder.

source/unix/opensslcrypto_ecc.c
DmitriyMusatkin: openssl is kinda inconsistent about those. those 3 frees are fine and can handle null (ex. https://github.com/aws/aws-lc/blob/0fed4dbc456406bdaf7158fd55d05a52d9998ce4/crypto/fipsmodule/ec/ec.c#L595 and similar in openssl).
but let me add null check just for consistency cause we usually check.

PR #243, opened 2025-11-26
Support static buffers in ecc signature helpers

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
current helper was relying on append byte dynamic that needs allocator. it should support buffer without allocator so switch which buffer function is being used.

### DIFF
--- a/source/ecc.c
+++ b/source/ecc.c
@@ -714,7 +714,7 @@ done:
 
 static struct aws_byte_cursor s_null_byte_cursor = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("\0");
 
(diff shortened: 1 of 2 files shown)


source/ecc.c
```diff
@@ -712,9 +712,11 @@ struct aws_ecc_key_pair *aws_ecc_key_new_from_hex_coordinates(
     return key;
 }
 
+static struct aws_byte_cursor s_null_byte_cursor = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("\0");
+
 int s_write_component_padded(struct aws_byte_buf *buf, struct aws_byte_cursor component, size_t pad_to) {
```
TingDaoK: trivial: static

TingDaoK (approved):

PR #251, opened 2026-07-16
Swap fatal for soft error

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*

currently ecdsa hard fails if coordinates do not match expected format. lets soft error instead of bringing process down.

### DIFF
--- a/source/windows/bcrypt_ecc.c
+++ b/source/windows/bcrypt_ecc.c
@@ -155,7 +155,9 @@ static int s_append_coordinate(
     size_t coordinate_size = aws_ecc_key_coordinate_byte_size_from_curve_name(curve_name);
     if (coordinate->len < coordinate_size) {
         size_t leading_zero_count = coordinate_size - coordinate->len;
-        AWS_FATAL_ASSERT(leading_zero_count + buffer->len <= buffer->capacity);
+        if (leading_zero_count + buffer->len > buffer->capacity) {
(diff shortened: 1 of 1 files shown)


TingDaoK (approved):

source/windows/bcrypt_ecc.c
```diff
@@ -155,7 +155,9 @@ static int s_append_coordinate(
     size_t coordinate_size = aws_ecc_key_coordinate_byte_size_from_curve_name(curve_name);
     if (coordinate->len < coordinate_size) {
         size_t leading_zero_count = coordinate_size - coordinate->len;
-        AWS_FATAL_ASSERT(leading_zero_count + buffer->len <= buffer->capacity);
+        if (leading_zero_count + buffer->len > buffer->capacity) {
+            return aws_raise_error(AWS_ERROR_CAL_MALFORMED_ASN1_ENCOUNTERED);
+        }
 
         aws_byte_buf_write_u8_n(buffer, 0x0, leading_zero_count);
```
TingDaoK, line 162: trivial: alternative check the return of `aws_byte_buf_write_u8_n`. it says 
```
 * If there is insufficient space in the buffer, returns false, leaving the
 * buffer unchanged.
```

DmitriyMusatkin: writing 0s to buffer will likely succeed, its the append after that fail. it does not really matter in this case my preference was to just not pollute buffer at all if it fails

PR #253, opened 2026-07-31
ECC Validation Test

sbSteveK:
Add a test that verifies `aws_ecc_key_pair_verify_signature` gracefully rejects a DER-encoded signature containing a coordinate larger than the curve's field size.

### DIFF
--- a/tests/ecc_test.c
+++ b/tests/ecc_test.c
@@ -1253,12 +1253,9 @@ AWS_TEST_CASE(ecc_key_gen_from_private_fuzz_test, s_ecc_key_gen_from_private_fuz
 #endif
 
 /*
- * A malformed DER-encoded ECDSA signature with an
- * oversized coordinate (70 bytes for P-256, which expects 32) must return an error
- * rather than triggering AWS_FATAL_ASSERT / abort().
- *
- * On Windows, the code path is in s_append_coordinate() in bcrypt_ecc.c.
- * On other platforms this exercises the equivalent DER parsing path.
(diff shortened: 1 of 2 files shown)


tests/ecc_test.c
```diff
@@ -1251,3 +1251,45 @@ static int s_ecc_key_gen_from_private_fuzz_test(struct aws_allocator *allocator,
...
+
+    struct aws_ecc_key_pair *key_pair = aws_ecc_key_pair_new_generate_random(allocator, AWS_CAL_ECDSA_P256);
+    ASSERT_NOT_NULL(key_pair);
+
+    /* 77-byte malformed P-256 DER signature: r = 70 bytes (oversized), s = 1 byte.
+     * For P-256, each coordinate should be at most 33 bytes (32 + 1 leading zero for sign). */
+    static const uint8_t malformed_sig[] = {
+        0x30, 0x4B,                                                 /* SEQUENCE, 75 bytes */
+        0x02, 0x46,                                                 /* INTEGER, 70 bytes (r) */
+        0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, /* r value: 70 bytes */
```
azkrishpy, line 1273: wouldn't the better test be if we showed r being 34 bytes is rejected? also, within this test can we also have a r is 33 bytes and s is 34 bytes case and show that fails too?

azkrishpy (approved):

tests/ecc_test.c
sbSteveK: https://github.com/awslabs/aws-c-cal/pull/254

PR #254, opened 2026-08-03
Expand ecc validation test

sbSteveK:
Add 33 and 34 byte tests to ecc validation test.


azkrishpy (approved):
We don't really need case 1 since case 2 covers the case of r exceeding 33 bytes.

